PR #8, opened 2019-05-28
Check in the initial version of an evaluation spec

shankari:
- initial spec
- initial notebook to experiment with filling out the spec
- basic setup and teardown scripts
- modified gitignore that excludes the copied over `conf` directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Copied over configuration files
+conf
+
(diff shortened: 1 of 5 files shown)


shankari:
This is actually to evaluate public data, and the related code is now in 
https://github.com/MobilityNet/mobilitynet-analysis-scripts

PR #9, opened 2019-06-12
Code to create, populate and validate an evaluation spec

shankari:


shankari:
This is actually for evaluating public data, and the resulting changes have already been checked in at 
https://github.com/MobilityNet/mobilitynet-analysis-scripts

PR #16, opened 2021-02-02
Bins and clusters

corinne-hcr:
Here is the bin trips on the map.
image

Here is the problem when running the `pipeline.cluster`
image


tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -0,0 +1,288 @@
...
+    "    trip_start_lat = trip.data.start_loc[\"coordinates\"][1]\n",
+    "    trip_start_log = trip.data.start_loc[\"coordinates\"][0]\n",
+    "    bin_start_lat.append(trip_start_lat)\n",
+    "    bin_start_log.append(trip_start_log)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "international-unemployment",
```
shankari, line 153: what is the difference between `international-unemployment` and `latest-cooling` cells?

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -0,0 +1,288 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "bin_end_lat = []\n",
```
shankari, line 157: where are you using these downstream?

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -0,0 +1,288 @@
...
+   "id": "lightweight-plaintiff",
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [],
+   "source": [
+    "# cluster the data using k-means\n",
+    "# def cluster(data, bins)\n",
+    "# return feat.clusters, feat.labels, feat.data\n",
+    "clusters,labels,feat_data = pipeline.cluster(bin_trips,bins)\n",
```
shankari, line 237: The error is fairly simple. The original pipeline calls `n, labels, data = cluster(data, len(bins))`.
You are calling `pipeline.cluster(bin_trips,bins)`.

So the original pipeline expects a number as the second argument, which is why it can multiply it by `1.5`
You are passing in a list, which is why you can't multiply it by a float.

shankari: if you had added some additional logging to `cluster(` you would have been able to discover this quickly!

shankari (changes requested):

tour_model_eval/trips_in_bins_and_clusters.ipynb
corinne-hcr: `bin_end_lat`, I use it to plot the lines on the map

tour_model_eval/trips_in_bins_and_clusters.ipynb
corinne-hcr: I don't know this. It is neither the code or output. Is it important?

shankari:
@corinne-hcr this works for me

```
--- a/tour_model_eval/trips_in_bins_and_clusters.ipynb
+++ b/tour_model_eval/trips_in_bins_and_clusters.ipynb
@@ -234,7 +234,7 @@
     "# cluster the data using k-means\n",
     "# def cluster(data, bins)\n",
     "# return feat.clusters, feat.labels, feat.data\n",
-    "clusters,labels,feat_data = pipeline.cluster(bin_trips,bins)\n",
+    "clusters,labels,feat_data = pipeline.cluster(bin_trips,len(bins))\n",
     "clusters"
    ]
   },
```

tour_model_eval/trips_in_bins_and_clusters.ipynb
corinne-hcr: Can you show an example of logging to figure out the what exactly the argument is?
`def cluster(data, bins)`is in the file right after she return bins in the previous function, that's why I pass bins

shankari:
@corinne-hcr I've made the change to call the cluster code. Do you need me to work on the plotting as well, or can you take it from here?

corinne-hcr:
OK, I think I can take it from here

tour_model_eval/trips_in_bins_and_clusters.ipynb
shankari: I'm trying to point out that you have two cells that seem to have the same code. The `"id"` is not really important, just to show you the cells that seem to be the same.

tour_model_eval/trips_in_bins_and_clusters.ipynb
shankari: Ah I apologize for missing that.

I don't understand why you are using this flow. It looks like you iterate through the bins and put the start and end points for all trips from all bins into one giant list. And then you plot all of them.

With this structure, how do you plan to use different colors for the different bins? Why don't you just iterate through the bins directly while plotting?

tour_model_eval/trips_in_bins_and_clusters.ipynb
shankari: can you add a link to where `def cluster(data, bins)` is?
Here's what I see
https://github.com/e-mission/e-mission-server/blob/master/emission/analysis/modelling/tour_model/cluster_pipeline.py#L87
which has `n, labels, data = cluster(data, len(bins))`

corinne-hcr: Oh, I didn't see the main when I call this function. 
[https://github.com/e-mission/e-mission-server/blob/master/emission/analysis/modelling/tour_model/cluster_pipeline.py#L61](url)

shankari:
I have a high level comment on your code structure which you need to address as well.
Let me know if I should add some sample code for that as well.

Otherwise, push early and often!

corinne-hcr:
I have a question, how can you modify the file that you haven't merged? Now I can change my file like the you do, but then I commit the file, there will be repeated modifications.

shankari:
you can pull from your branch (`origin <branch>`)

tour_model_eval/trips_in_bins_and_clusters.ipynb
shankari: @corinne-hcr here is how you would clarify that function + add additional logs
https://github.com/e-mission/e-mission-server/pull/792

corinne-hcr:
I can't plot the cluster graph. I can't access the points through the pipeline.cluster, I have to run `featurization.featurization(bin_trips)`to get the points.` pipeline.cluster`doesn't return`points` Also, I have questions about labels. 
`[1,
 1,
 1,
 1,
 1,
 5,
 5,
 5,
 5,
 5,
 6,
 6,
 6,
 6,
 6,
 2,
 2,
 3,
 3,
 0,
 0,
 4,
 4,
 8,
 8,
 7,
 7]`Since I don't see other sort method, does it mean the five trips from the beginning belong to label 1?

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -130,138 +140,184 @@
...
    },
    "outputs": [],
    "source": [
-    "# bin_trips"
+    "# Plot all the bin trips on the map, later we should use different color for different bins, but not sure whether it works.\n",
+    "m = folium.Map(location=[bin_trips[0].data.start_loc[\"coordinates\"][1], bin_trips[0].data.start_loc[\"coordinates\"][0]],zoom_start=12, max_zoom= 30, control_scale=True)\n",
+    "t_index = 0\n",
+    "for bin in bins:    \n",
+    "    for i in range (len(bin)):\n",
+    "        layer = folium.PolyLine([[bin_trips[t_index].data.start_loc[\"coordinates\"][1],bin_trips[t_index].data.start_loc[\"coordinates\"][0]],\n",
```
shankari, line 176: can you add a comment here about why you are using `t_index` instead of `i` for the plotting?
I can figure it out (you want all the trips in the same bin to be the same color eventually) but as you have found from reading Naomi's code, it is very helpful to have comments in the code that explain your thought process to the next reader.

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -130,138 +140,184 @@
...
+    "                                [bin_trips[t_index].data.end_loc[\"coordinates\"][1],bin_trips[t_index].data.end_loc[\"coordinates\"][0]]],weight=2, color='orange')\n",
```
shankari, line 177: why are you not changing the color?

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -130,138 +140,184 @@
...
-    "bin_start_lat = []\n",
-    "bin_start_log = []\n",
-    "\n",
-    "# Flip indices because points are in geojson (i.e. lon, lat)\n",
-    "for trip in bin_trips:\n",
-    "    trip_start_lat = trip.data.start_loc[\"coordinates\"][1]\n",
-    "    trip_start_log = trip.data.start_loc[\"coordinates\"][0]\n",
-    "    bin_start_lat.append(trip_start_lat)\n",
-    "    bin_start_log.append(trip_start_log)"
+    "# bin_trips"
```
shankari, line 149: why do you have a cell whose entire code is just a commented out line for `bin_trips`?

shankari (changes requested):
I will make most of the changes myself, but I would like you to add the comments.
It is good practice :)

Also, please figure out how to use different colors for each bin.

tour_model_eval/trips_in_bins_and_clusters.ipynb
corinne-hcr: Oh, codes that are commented out are the way I try to see the output of the result like`bin_trips`,`bins` etc. They don't really necessary in the code.

tour_model_eval/trips_in_bins_and_clusters.ipynb
corinne-hcr: Actually, from Naomi's code and the output, the `bin_trips`is filtered data, but they don't follow the index in bins. I think she just appended the filtered data in the new data list, so they don't follow `i`.

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -0,0 +1,344 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "occasional-basis",
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [],
+   "source": [
+    "# This doesn't show the map, I don't know why.\n",
+    "map_clusters(feat)"
```
shankari, line 301: because you weren't returning the cluster map.
In the previous viz, you didn't have a function to display the map, you did it directly in a cell
so typing out `m` at the end was enough

But here, you are calling a function. The function needs to return the map otherwise the calling cell doesn't get it.
Pretty straightforward.

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -0,0 +1,344 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "round-death",
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "bin_trips, bins = pipeline.remove_noise(trips, radius)"
```
shankari, line 137: We should revisit the level at which we are calling the cluster pipeline functions for visualization.
For example, `remove_noise` doesn't only bin the values, it also removes all the uncommon bins.
So we can't visualize/work with those bins because we never see them.

That is fine if we agree with the novelty detection method; not so great if we don't.
Similar considerations apply to calling `cluster` versus the featurization code directly

tour_model_eval/trips_in_bins_and_clusters.ipynb
corinne-hcr: Oh I see!!!
But I still don't know how to plot the map based on labels? `clusters`after running `pipeline.cluster`is just the number of cluster. But the points are in a list with their own orders. How can they connect to the labels?

shankari: Correct, this is a necessary but not a sufficient fix. Please see my next comment for the level at which you need to work with the code.

tour_model_eval/trips_in_bins_and_clusters.ipynb
shankari: In that case, you should remove them from the notebook before submitting for review.

tour_model_eval/trips_in_bins_and_clusters.ipynb
shankari: I am not sure what you mean by that. Can you give an example? And maybe add the example to the comment as well.

shankari:
@corinne-hcr please check my commit. Both the bin and the cluster displays work, both before and after deleting novel bins.
Please use this to figure out the difference between bins and clusters before the 10am meeting tomorrow.
You may want to visualize for all users - I think that the user you have selected has nClusters == nBins

```
DEBUG:root:number of bins after filtering: 8
```

```
DEBUG:root:number of clusters: 8
```

Again, you are encouraged to add log statements to the underlying server code to understand it better.

corinne-hcr:
I have some questions. 
1.In `Trips from all bins`, you call `similarity.similarity(trips, radius)` and `sim.bin_data()`. It shows `DEBUG:root:After removing trips that are points, there are 51 data points`

But in `Trips from bins above the cutoff only`, That was the way I called `pipeline.remove_noise(trips, radius)`. It shows `DEBUG:root:After removing trips that are points, there are 50 data points`The number of data points is different. Shouldn't it be the same number of data points?

2.After calling `pipeline.remove_noise(trips, radius)`, it shows
`the new number of trips is 22`
`the cutoff point is 8`
`number of bins after filtering: 8`

But in my previous code, I remember I had around 23 trips, but the cutoff point was 9, and there were 9 bins.
That also shows in the`viz_similarity_unlabeled.py`
image
There are 9 bins above the cutoff point.

I don't know how that affect the outcome.

shankari:
@corinne-hcr

Great questions! I would encourage you to experiment with the code (potentially by adding additional log statements) or running individual code snippets to resolve the discrepancies.

wrt answers, I don't know either. Again, I want to reiterate that this is not a class. I don't know the answers. You have to find out the answers and tell me.

corinne-hcr:
Outcome:

1. Due to unknown reason, I test `bins/clusters on all data` and `bins/clusters above cutoff points` separately. I also change a dataset with more data.
2. Regarding the original `min_cluster` = 2, I find some materials online that also start with k=2. I think that clustering means at least it divide one big group into two clusters.
3. For `max_cluster`, she used `1.5 * bins`. I think that might be an empirically setting, but based on the implementation, the final number of clusters never hit the max. 
4. For `kmedoids`, I think we can delete it. We never use this in implementation, I think she just left an option here.
5. About bins and cluster

- on all data
` DEBUG:root:There are 88 bins before filtering`
`DEBUG:root:number of clusters: 33`
Bins are based on the distance of start points and end points of two trips. Clusters are based on silhouette score. 
I add some logs on the server code. Here is how silhouette score works to determine the clusters.
image
So, we can see here, (the original max she set was 2) if sil > max, max will be set to equal to sil. As tests keep going, the sil reach a peak, and then decrease. When cluster = 33 is the peak of sil score. That how sil score finds the best cluster for all data.
There is `r = max_clusters - min_clusters+1`. Since min_cluster = 2, max_clusters = 1.5*bins, r is the test times from 2 to max_clusters.

Here is the trip labels after clustering
image


- on data above cutoff only

For bins
`DEBUG:root:After removing trips that are points, there are 154 data points`
`DEBUG:root:number of bins before filtering: 88`
`DEBUG:root:the new number of trips is 52`
`DEBUG:root:the cutoff point is 6`
`DEBUG:root:number of bins after filtering: 6`

`DEBUG:root:The list of bins is [[9, 14, 23, 25, 32, 34, 39, 53, 67, 69, 71, 87, 92, 112, 114, 128, 149, 151, 153], [10, 15, 24, 26, 33, 35, 40, 58, 68, 70, 88, 93, 113, 115, 150, 152], [5, 18, 77, 80, 142], [11, 16, 74, 131], [22, 64, 109, 130], [52, 61, 111, 134]]`


For clusters
Instead of following the original code `min = bins`, I set `min = 0`, so I can see if sil score also finds the same number of clusters.

Here is what I have
`DEBUG:root:number of clusters: 6`

image

So, given the same amount of trips, sil score also find the same number of clusters, and the trip labels distribution is exact the same as the list of bins (same order and same number of trips in one cluster/bin).

- Why she used clustering after binning
I think the one of reasons could be that sometimes elbow method is hard to find the best k. Some online materials state that sil score is better and more straight forward to find the best k. Here she put common trips bins into cluster can examine if she got the best clusters.

-Why not just use clustering
I think, the problem is, I don't see a method to find common trips after clustering(if we put all data into analysis, how to find 6 common trips clusters out of 33). The elbow method in clusters is just to find the k in given data. So that she used cutoff point in bins can be considered a good way to find common trips.


Next, I am going to test on different users and see how all data result in bins and cluster. Also, I will visualize some part of the bins and clusters.

shankari:
@corinne-hcr this is very good. Couple of clarifications:
- "I also change a dataset with more data." I think you mean that you changed to a user with more data. A dataset typically contains data from multiple users. So a new dataset would be data from a completely different source.
- "For max_cluster, she used 1.5 * bins. I think that might be an empirically setting, but based on the implementation, the final number of clusters never hit the max." Again, would be good to verify at least against all the users we have before making that bold statement.
- "For kmedoids, I think we can delete it. We never use this in implementation, I think she just left an option here." Great that she left the option there. That allows us to compare bins v/s clusters, by actually evaluating bins v/s kmeans clusters v/s kmedoid clusters
- "So, given the same amount of trips, sil score also find the same number of clusters." As we discussed, you should verify this finding against multiple users as well. In particular, when I integrated Naomi's code into the system, I found that there were differences sometimes https://github.com/e-mission/e-mission-server/commit/d6a41c681d2eb4bcde983e0056128dcc97d99de0
- "I think, the problem is, I don't see a method to find common trips after clustering(if we put all data into analysis, how to find 6 common trips clusters out of 33). " I think what you are saying is that clustering will put all the trips into one of the k clusters. So we will not be able to split into common and novel. Good finding! 👏

shankari:
As you have outlined, the next tasks are:
- Test on different users and see how all data result in bins and cluster.
- Also, I will visualize some part of the bins and clusters and see if they match 1:1 in terms of trips
  - visualizing on a map with different colors will help if we are considering 1 or 2 users
  - for an overall picture, you might want to actually compare the trip ids in the bins and the clusters to determine whether the bins and clusters are identical or not

shankari:
"I think, the problem is, I don't see a method to find common trips after clustering(if we put all data into analysis, how to find 6 common trips clusters out of 33)." 

Given this, does it matter that we perform steps in the pipeline order by binning and then clustering. If binning and clustering are largely independent, would it work to cluster first, use the silhouette score to determine k and choose that as the elbow?

corinne-hcr:
Update:
Test on different users and see how all data result in bins and cluster. But I will actually put all data and above cutoff data results here since I notice there is something we can discuss.

user1:
-all data
`DEBUG:root:There are 110 bins before filtering`
`DEBUG:root:number of clusters: 14`
-above cutoff only
`DEBUG:root:number of bins after filtering: 16`
`DEBUG:root:number of clusters: 16`

user2:
-all data
`DEBUG:root:There are 37 bins before filtering`
`DEBUG:root:number of clusters: 8`
-above cutoff only
`DEBUG:root:number of bins after filtering: 9`
`DEBUG:root:number of clusters: 9`

user3:
-all data
`DEBUG:root:There are 143 bins before filtering`
`DEBUG:root:number of clusters: 62`
-above cutoff only
`DEBUG:root:number of bins after filtering: 14`
`DEBUG:root:number of clusters: 13`
Note that there are actually 14 cluster labels 
`[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 11, 11, 11, 11, 11, 11, 0, 0, 0, 0, 0, 8, 8, 8, 8, 8, 5, 5, 5, 5, 5, 10, 10, 10, 10, 7, 7, 7, 7, 4, 4, 4, 9, 9, 9, 2, 2, 2, 6, 6, 6, 12, 12, 12]`  The sil score decreases from 14 clusters
```DEBUG:root:testing 13 clusters
DEBUG:root:sil is 0.950153323026658
DEBUG:root:sil > max True
DEBUG:root:sil is 0.950153323026658, max is 0.9283151017642491
DEBUG:root:The new max is 0.950153323026658
DEBUG:root:testing 14 clusters
DEBUG:root:sil is 0.9422369415328851
```

user4:
-all data
`DEBUG:root:There are 32 bins before filtering`
`DEBUG:root:number of clusters: 13`
-above cutoff only
`DEBUG:root:number of bins after filtering: 2`
`DEBUG:root:number of clusters: 2`

user5:
-all data
`DEBUG:root:There are 196 bins before filtering`
`DEBUG:root:number of clusters: 2`
-above cutoff only
`DEBUG:root:number of bins after filtering: 17`
`DEBUG:root:number of clusters: 17`

user6:
-all data
`DEBUG:root:There are 71 bins before filtering`
`DEBUG:root:number of clusters: 8`

-above cutoff only
`DEBUG:root:number of bins after filtering: 9`
```DEBUG:root:The list of bins is [[1, 3, 12, 17, 23, 25, 32, 39, 41, 44, 47, 72, 104, 108], [4, 16, 19, 24, 26, 33, 40, 43, 46, 48, 78, 103, 107, 110], [28, 45, 100, 101, 102, 105, 106, 109, 123], [27, 75, 79, 82, 99, 113, 122], [2, 14, 81, 83, 98], [13, 42, 80, 120], [7, 20, 30], [9, 35, 62], [73, 77, 114]]```

`DEBUG:root:number of clusters: 6`
```[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 5, 5, 5, 5, 5, 5, 5, 1, 1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 4, 4, 4, 2, 2, 2]```

I think here we should plot it in map to see why different bins would be assigned same labels(clusters)
Same as user 3

user7:
-all data
`DEBUG:root:There are 88 bins before filtering`
`DEBUG:root:number of clusters: 33`
-above cutoff only
`DEBUG:root:number of bins after filtering: 6`
`DEBUG:root:number of clusters: 6`

user8:
-all data
`DEBUG:root:There are 268 bins before filtering`
`DEBUG:root:number of clusters: 4`
-above cutoff only
`DEBUG:root:number of bins after filtering: 46`
`DEBUG:root:number of clusters: 44`

user9:
-all data
`DEBUG:root:There are 164 bins before filtering`
`DEBUG:root:number of clusters: 87`
-above cutoff only
`DEBUG:root:number of bins after filtering: 27`

`DEBUG:root:number of clusters: 26`
```[0, 0, 0, 0, 0, 0, 0, 7, 7, 7, 7, 7, 7, 2, 2, 2, 2, 2, 2, 6, 6, 6, 6, 6, 6, 12, 12, 12, 12, 12, 12, 14, 14, 14, 14, 14, 14, 21, 21, 21, 21, 21, 13, 13, 13, 13, 13, 1, 1, 1, 1, 19, 19, 19, 19, 20, 20, 20, 24, 24, 24, 16, 16, 16, 3, 3, 3, 17, 17, 5, 5, 4, 4, 18, 18, 9, 9, 15, 15, 6, 6, 23, 23, 10, 10, 25, 25, 22, 22, 11, 11, 8, 8]``` 
See label `6`, same thing here

user 10:
-all data
`DEBUG:root:There are 102 bins before filtering`
`DEBUG:root:number of clusters: 35`
-above cutoff only
`DEBUG:root:number of bins after filtering: 21`
`DEBUG:root:number of clusters: 21`

user11:
-all data
`DEBUG:root:There are 132 bins before filtering`
`DEBUG:root:number of clusters: 27`
-above cutoff only
`DEBUG:root:number of bins after filtering: 14`
```DEBUG:root:The list of bins is [[12, 22, 49, 51, 53, 63, 72, 175], [13, 25, 38, 54, 73, 139, 144], [24, 35, 37, 90, 127, 143, 172], [7, 17, 27, 29, 118, 165], [78, 111, 133, 150, 201], [8, 28, 119, 166], [23, 36, 52, 126], [32, 87, 132, 146], [56, 92, 97, 117], [112, 134, 151, 202], [113, 182, 189, 195], [26, 140, 145], [50, 62, 64], [131, 154, 161]]```

`DEBUG:root:number of clusters: 8`
```[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 2, 2, 2, 2, 1, 1, 1, 1, 5, 5, 5, 5, 6, 6, 6, 6, 0, 0, 0, 0, 2, 2, 2, 2, 4, 4, 4, 1, 1, 1, 7, 7, 7]```

user12 skip

user13:
-all data
`DEBUG:root:There are 303 bins before filtering`
`DEBUG:root:number of clusters: 185`
-above cutoff only
`DEBUG:root:number of bins after filtering: 11`
`DEBUG:root:number of clusters: 11`


As we can see from the output above, bins are more sensitive than clusters. If we directly put all data in cluster then use elbow method, we may not find the common trips as expected. Of course, I need to plot some special cases where bins are significantly different from cluster.

Second concern, for data above cutoff, there are some example show that `bin_trips` clusters are less than the number of `bins`. I need to plot those on map to see if bins wrongly put similar trips into different bins or clusters incorrectly label the trips.(Maybe I will just need to plot those trips have same label but in different bins)

**Comment by shankari** [comment]
all that sounds good. My only other suggestion would be to capture the values here {all trips: bins, cluster, above cutoff: bins, cluster} in a data structure (maybe a simple data frame) and plot it. We can then include that plot in the report along with the maps that visualize the discrepancies (user3, user6,...)

**Comment by shankari** [comment]
also, don't forget to push up your changes to the PR after that is done!

**Comment by corinne-hcr** [comment]
Do you mean use pandas data frame to show the values?({all trips: bins, cluster, above cutoff: bins, cluster})
Do you need the notebook now to see the difference? I didn't make many modifications. Maybe I push it later when I done with the map?

**Comment by shankari** [comment]
Yes, you should push after you collate the values and plot maps for the outliers
"pandas data frame to show the values" you can even plot them using `DataFrame.plot`!

**Comment by corinne-hcr** [comment]
Partially update: the difference between trips in the same bin and trips in the same cluster

1. On all data
Here I choose user2 as an example
`DEBUG:root:There are 37 bins before filtering`
`DEBUG:root:number of clusters: 8`

Here is the map for bins
image



Here is the map for clusters. 
Screen Shot 2021-02-08 at 9 06 09 AM

As we can see, in the cluster (label 1), the trips are largely different (red lines). However, those trips are in different bins. On all data, bins make more sense than clusters.


2. On above cutoff only
Here I choose user 11 as an example
`DEBUG:root:number of bins after filtering: 14`
`DEBUG:root:number of clusters: 8`

Here is the bins list
```DEBUG:root:The list of bins is [[12, 22, 49, 51, 53, 63, 72, 175], [13, 25, 38, 54, 73, 139, 144], [24, 35, 37, 90, 127, 143, 172], [7, 17, 27, 29, 118, 165], [78, 111, 133, 150, 201], [8, 28, 119, 166], [23, 36, 52, 126], [32, 87, 132, 146], [56, 92, 97, 117], [112, 134, 151, 202], [113, 182, 189, 195], [26, 140, 145], [50, 62, 64], [131, 154, 161]]```

I will compare the trips in the first 3 bins and the last 2 bins with trips in label 1

Here is the trips from the first 3 bins
Screen Shot 2021-02-08 at 9 35 51 AM


Here is the trips from the last 2 bins
image


Here is the label list
```[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 6, 6, 6, 6, 6, 6, 3, 3, 3, 3, 3, 6, 6, 6, 6, 1, 1, 1, 1, 5, 5, 5, 5, 2, 2, 2, 2, 0, 0, 0, 0, 4, 4, 4, 1, 1, 1, 7, 7, 7]```

Here is the trips in label 1
image

**Comment by corinne-hcr** [comment]
update the graph of bins and clusters (from data above cutoff) :
image
image

File: tour_model_eval/trips_in_bins_and_clusters.ipynb [hunk 9]
```diff
@@ -42,293 +49,789 @@
...
    "metadata": {},
    "outputs": [],
    "source": [
-    "# we should experiment with different values here\n",
-    "radius = 300"
+    "# Plot all the bin trips on the map, use different color for different bins.\n",
+    "m = folium.Map(location=[trips[0].data.start_loc[\"coordinates\"][1], trips[0].data.start_loc[\"coordinates\"][0]],zoom_start=12, max_zoom= 30, control_scale=True)\n",
+    "for t in range (len(sim.bins)):    \n",
+    "    for i in range (len(sim.bins[t])):\n",
+    "        layer = folium.PolyLine([[trips[sim.bins[t][i]].data.start_loc[\"coordinates\"][1],trips[sim.bins[t][i]].data.start_loc[\"coordinates\"][0]],\n",
```
shankari, line 158: can you add some comments here on how you are selecting trips to help future students? I would recommend pulling out the current trip selection into a variable so you don't repeat yourself. You should also use `for bin in bins` instead of the index unless you actually need the index, so 

```
for t, curr_bin in sim.bins:
   for curr_trip_index in curr_bin:
      curr_trip = trips[curr_trip_index]
....
```

Much easier to understand and less convoluted

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -42,293 +49,789 @@
...
+    "m.add_child(cmp_all_bins)     \n",
+    "m"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "better-continent",
+   "metadata": {},
+   "source": [
+    "### Trips from bins above the cutoff only"
```
shankari, line 171: As I said, please move this to a separate notebook. In general, it is good to follow the [Golden Rule](https://en.wikipedia.org/wiki/Golden_Rule) while writing code - think about the level of clarity and commenting that you would have liked to see in Naomi's code, and introduce it into your code 🙏

Think about the next person who will work on this code and have questions like "what is this supposed to be doing?"

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -42,293 +49,789 @@
...
-    "        t_index += 1\n",
-    "        layer.add_to(m)\n",
-    "        \n",
-    "m"
+    "#Plot the clusters above the cutoff only with different colors,to see if the algorithm wrongly puts different trips into the same cluster\n",
+    "cluster_cutoff = folium.Map(location=[feat.points[0][1], feat.points[0][0]],zoom_start=12, max_zoom= 30, control_scale=True)\n",
+    "\n",
+    "print(feat.labels)\n",
+    "if feat.labels:             \n",
+    "    for i in range(len(feat.points)):\n",
```
shankari, line 415: again, can use `for i, point in enumerate(feat.points)` or even `for color, point in zip(color, points)`

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -42,293 +49,789 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "second-instrumentation",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# We use user11 as an example\n",
+    "#-above cutoff only\n",
+    "#DEBUG:root:number of bins after filtering: 14\n",
```
shankari, line 466: can we remove these log statements? Why are they useful?
Isn't this it the same as including outputs in the notebook?

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -42,293 +49,789 @@
...
-    "    print(self.labels)\n",
-    "    if self.labels:             \n",
-    "        for i in range(len(self.points)):\n",
-    "            start_lat = self.points[i][1]\n",
-    "            start_lon = self.points[i][0]\n",
-    "            end_lat = self.points[i][3]\n",
-    "            end_lon = self.points[i][2]\n",
-    "            path = [(start_lat, start_lon), (end_lat, end_lon)]\n",
+    "print(feat.labels)\n",
+    "if feat.labels:             \n",
```
shankari, line 627: I would pull this out into a separate `.py` module to use it in both notebooks.

tour_model_eval/trips_in_bins_and_clusters.ipynb
```diff
@@ -42,293 +49,789 @@
...
   },
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "about-equity",
+   "id": "employed-baltimore",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "user_index = ['user1','user2','user3','user4','user5','user6','user7','user8','user9','user10','user11','user13']"
```
shankari, line 773: why does this user_index have only a subset of users? Where is user12?

shankari (changes requested):
Can you please clean up the code and resubmit?

tour_model_eval/trips_in_bins_and_clusters.ipynb
corinne-hcr: user12 doesn't have data above cutoff. 
image

tour_model_eval/viz_bin_cluster.py
```diff
@@ -0,0 +1,122 @@
...
+            end_lat = point[3]
+            end_lon = point[2]
+            layer = folium.PolyLine([[start_lat, start_lon],
+                                     [end_lat, end_lon]], weight=2, color=color_map(labels[i]))
+            layer.add_to(map)
+    map.add_child(color_map)
+    return map
+
+
+def specific_bin_map(bins, trips):
```
shankari, line 80: Isn't this the same as `bins_map` but by passing in the specific bins as the parameter?
Can merge them!

tour_model_eval/viz_bin_cluster.py
```diff
@@ -0,0 +1,122 @@
...
+                [[curr_trip.data.start_loc["coordinates"][1], curr_trip.data.start_loc["coordinates"][0]],
+                 [curr_trip.data.end_loc["coordinates"][1], curr_trip.data.end_loc["coordinates"][0]]], weight=2,
+                color=color_map(t))
+
+            layer.add_to(map)
+    map.add_child(color_map)
+    return map
+
+
+def specific_cluster_map(labels, points, clusters):
```
shankari, line 101: isn't this just a duplicate of `clusters_map` with a different set of clusters passed in?
Again, can merge.
DRY.

tour_model_eval/viz_bins_clusters_above_cutoff.ipynb
```diff
@@ -0,0 +1,392 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "informative-mitchell",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "logging.debug('There are %s users, we will plot the graphs for one of them' % len(all_users))\n",
+    "\n",
+    "user = all_users[1]"
```
shankari, line 61: Is there a reason we are picking the second user?
Might want to add some comments around this choice.

tour_model_eval/viz_bins_clusters_above_cutoff.ipynb
```diff
@@ -0,0 +1,392 @@
...
+   "source": [
+    "## Visualizing clusters above cutoff"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "played-gabriel",
+   "metadata": {},
+   "source": [
+    "Note: should run \"Trips from bins above the cutoff only\" first"
```
shankari, line 136: Remove note since it is not relevant any more.

tour_model_eval/viz_bins_clusters_above_cutoff.ipynb
```diff
@@ -0,0 +1,392 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "frozen-modern",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# copied from the cluster pipeline code so we can work with the featurization code directly\n",
```
shankari, line 146: since this is above the cutoff, I think we can just use the `pipeline.cluster` function directly. I don't see us using the `feat` object later, only `feat.clusters, feat.labels, feat.data` which is what the `pipeline.cluster` function returns.

tour_model_eval/viz_bins_clusters_above_cutoff.ipynb
```diff
@@ -0,0 +1,392 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "freelance-greensboro",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# user12 is not typical, so we filter it out"
```
shankari, line 286: convert this to markdown?

tour_model_eval/viz_bins_clusters_above_cutoff.ipynb
```diff
@@ -0,0 +1,392 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "spare-worst",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "user_index = ['user1','user2','user3','user4','user5','user6','user7','user8','user9','user10','user11','user13']"
```
shankari, line 336: Add a comment on you are doing this to help others later.

tour_model_eval/viz_bins_clusters_all_data.ipynb
```diff
@@ -0,0 +1,332 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "corresponding-silly",
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "# copied from the cluster pipeline code so we can work with the featurization code directly\n",
+    "feat = featurization.featurization(trips)\n",
```
shankari, line 156: here, it makes sense to use more of a white box model

tour_model_eval/viz_bins_clusters_all_data.ipynb
```diff
@@ -0,0 +1,332 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "ultimate-bottom",
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "sim = similarity.similarity(trips, radius)\n",
```
shankari, line 102: and here too.

shankari (changes requested):
Very good refactoring! Some more improvements and then we are good to go.

tour_model_eval/viz_bin_cluster.py
corinne-hcr: No, I don't think they are the same. Since we can choose one or some specific bin(s) here, we highly rely on the indices. In my example, I choose 3 bins to compare with 1 cluster.

tour_model_eval/viz_bin_cluster.py
corinne-hcr: They look like similar. But I add a if statement in this function. If I merge them, then other people might need to fix the indentation every time and they couldn't run all at a time. I don't know how to avoid this situation .

tour_model_eval/viz_bin_cluster.py
shankari: I don't understand this. Why can't you pass in a subset of the bins to `all_bin_map`? e.g. `sim.bins[0:3]` should give you first three bins.

tour_model_eval/viz_bins_clusters_all_data.ipynb
shankari: no changes are needed, this is really just a comment.

tour_model_eval/viz_bins_clusters_all_data.ipynb
shankari: no changes are needed, this is really just a comment.

tour_model_eval/viz_bin_cluster.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # Plot all clusters with different colors on the map
+    # Choose the first start point to locate the map
+    map = folium.Map(location=[points[0][1], points[0][0]], zoom_start=12, max_zoom=30,
+                             control_scale=True)
+
+    if labels:
+        for i, point in enumerate(points):
+            # Here we can choose a specific cluster to plot
+            if labels[i] == 1:
```
shankari, line 113: ah I see, this is the `if`. The way to resolve that is to again pass in a parameter. without that change, users will have to edit the python file every time they want to change the cluster that is plotted. Instead, you should merge the functions and then if you want to plot a particular cluster, you pass in the selected label or `None` if you want to plot all labels.

tour_model_eval/viz_bin_cluster.py
corinne-hcr: Oh ok!

tour_model_eval/viz_bin_cluster.py
shankari: I would actually go further and pass in a list of selected labels and change the `if` to `if sel_label_list is None or labels[i] in sel_label_list`

tour_model_eval/viz_bin_cluster.py
```diff
@@ -15,108 +15,86 @@
...
-def bins_map(bins, trips):
+def bins_map(bins, bins_list, trips):
     # Plot all the bin trips on the map, use different colors for different bins.
     # Each color represents one bin based on index
     # Choose the first start point from the trips to locate the map for convenience
-    color_map = cm.linear.Set1_07.to_step(len(bins),index=[i for i in range (len(bins))])
+
+    # The function takes three parameters:
+    # - bins: bins list above cutoff after running pipeline.remove_noise
+    # - bins_list: the list of indices of selected bins
```
shankari, line 25: nit: rename to `sel_bin_indices`?

tour_model_eval/viz_bin_cluster.py
```diff
@@ -15,108 +15,86 @@
...
+                    [[curr_trip.data.start_loc["coordinates"][1], curr_trip.data.start_loc["coordinates"][0]],
+                     [curr_trip.data.end_loc["coordinates"][1], curr_trip.data.end_loc["coordinates"][0]]], weight=2,
+                    color=color_map(index))
+                layer.add_to(map)
+    map.add_child(color_map)
+    return map
 
 
-def clusters_map(labels, points, clusters):
+def clusters_map(labels,labels_list, points, clusters):
```
shankari, line 62: Ditto.

tour_model_eval/viz_bin_cluster.py
```diff
@@ -15,108 +15,86 @@
...
-
-            layer.add_to(map)
-    map.add_child(color_map)
-    return map
-
-
-def specific_cluster_map(labels, points, clusters):
+    # The function takes four parameters:
+    # - labels: trip label list after running featurization/pipeline.cluster
+    # - labels_list: the list of selected label
```
shankari, line 71: nit: rename with `sel_label_list`?

tour_model_eval/viz_bins_clusters_above_cutoff.ipynb
```diff
@@ -3,7 +3,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "contemporary-bennett",
+   "id": "proud-recruitment",
```
shankari, line 6: nit: why are all these IDs different?

tour_model_eval/viz_bin_cluster.py
```diff
@@ -15,108 +15,86 @@
 import branca.colormap as cm
 
 
-def bins_map(bins, trips):
+def bins_map(bins, bins_list, trips):
```
shankari, line 18: why can't you just pass in the bins that are selected by the `bins_list`?
having an `if` and `else` inside the function is still messy and I am not convinced that they are needed.

e.g. http://stackoverflow.com/questions/19252301/ddg#33356141

tour_model_eval/viz_bin_cluster.py
```diff
@@ -15,108 +15,86 @@
...
+            for i, point in enumerate(points):
+                if labels[i] == label:
+                    start_lat = point[1]
+                    start_lon = point[0]
+                    end_lat = point[3]
+                    end_lon = point[2]
+                    layer = folium.PolyLine([[start_lat, start_lon],
+                                             [end_lat, end_lon]], weight=2, color=color_map(labels[i]+1))
+                    layer.add_to(map)
+    else:
```
shankari, line 89: this is essentially the same as the two function solution before, with the part in for loop being almost identical.
are the labels a list? can't you filter the selected label list similar to the bins above?

shankari (changes requested):

tour_model_eval/viz_bins_clusters_above_cutoff.ipynb
corinne-hcr: I don't know. Is it due to change of network? checkpoints in Jupiter notebook?

tour_model_eval/viz_bin_cluster.py
corinne-hcr: I can create a `new_bins_list` that contains the selected bins, and let `bins` = `new_bins_list` in order to reuse the loop for general bins (in order, not selected). But then when I run the loop, it relies on the index in `bins` to plot the map. For example, if I pass[1,7,8], then `new_bins_list = [bins[1],bins[7],bin[8]]`, but the indices for `bins[1],bins[7],bin[8]` would be `0,1,2`. Then the map would use the first, second, and third color to plot (which are designed for the original bin index 0,1,2). That is different from the colormap, I think others will be confused.
The reason I need to pass in the bins is that I use it to fixed the colormap. Because I am actually using  `StepColormap` to make each bin has its own color. The color map highly relies on the indices. From the example above, if the indices is 0,1,2, I need to make the indices for colormap from 0 to 3 in order to have 3 colors, but in the colormap it will just show 0 1 2 3, it will not show 1 7 8.
Though the code looks repeated, but I think they will messy if I don't separate them. What do you think?

tour_model_eval/viz_bin_cluster.py
corinne-hcr: `labels`is a list like`[1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 1,
 2,
 2,
 2,
 2,
 2,
 2,
 3,
 3,
 3,
 3,
 3,
 2,
 2,
 2,
 2,
 1,
 1,
 1,
 1,
 5,
 5,
 5,
 5,
 6,
 6,
 6,
 6,
 0,
 0,
 0,
 0,
 2,
 2,
 2,
 2,
 4,
 4,
 4,
 1,
 1,
 1,
 7,
 7,
 7]`, I will need to compare the label for each trip with the selected labels, and only plot the points have selected labels. For general clusters, I just need to plot all the points. I don't see the possibility to reduce the code here, unless create another function for the same part of code.
Also, I need to pass in the labels to determine the color map.

tour_model_eval/viz_bin_cluster.py
shankari: the answer is simple and one I had hinted before. Associate the bins and colors before subsetting them.
`zip(bins,color)` will give you `(bin, color)` tuples - 

```
>>> colors = ["white", "gray", "green", "yellow"]
>>> bins = ["b1", "b2", "b3", "b4"]
>>> list(zip(bins, colors))
[('b1', 'white'), ('b2', 'gray'), ('b3', 'green'), ('b4', 'yellow')]
>>> list(zip(bins, colors))[2:4]
[('b3', 'green'), ('b4', 'yellow')]
>>> list(zip(bins, colors))[1:3]
[('b2', 'gray'), ('b3', 'green')]
>>>
```

Now, you can do:

```
for bin, color in sel_bin_colors:
....
```

tour_model_eval/viz_bin_cluster.py
shankari: why can't your `if` then be

```
if sel_labels is None or trip.label in sel_labels:
    # plot point
```

tour_model_eval/viz_bin_cluster.py
corinne-hcr: But my color is not a list like `colors = ["white", "gray", "green", "yellow"]`, I don't know what colors in the colormap, it depends on the steps/labels

shankari: This is an example of how zip can work. The bins are also not "b1", "b2", "b3"...

It is a list of some colors, right? You can only index it because it is a list. 

You can zip your color list (whatever it is) with the bin list. Did you try it?

tour_model_eval/viz_bin_cluster.py
```diff
@@ -15,51 +15,40 @@
...
     # - bins: bins list above cutoff after running pipeline.remove_noise
-    # - bins_list: the list of indices of selected bins
+    # - sel_bin_indices: the list of indices of selected bins (set to None for plotting all bins)
     # - trips: the trips after running pipeline.read_data
 
     # In order to plot trips correctly, map index should be len(bins)+1,
     # so that the color steps length is equal to the bins number. For bin index which is 0,
     # will be plotted in color which is in index 1.
-    color_map = cm.linear.Set1_07.to_step(len(bins)+1, index=[i for i in range(len(bins)+1)])
+    color_map = cm.linear.Set1_07.to_step(len(bins), index=[i for i in range(len(bins)+1)])
```
shankari, line 31: I am not going to force you to change this, but I want to point out that you can in fact get colors out of a colormap object. In fact, there are two ways to do so:

First, we create the colormap

```
>>> import branca.colormap as cm
>>> color_map = cm.linear.Set1_07.to_step(10, index=range(10))
>>> color_map
<branca.colormap.StepColormap object at 0x7f91c8356fd0>
```

Option 1: create a list through iteration.

```
>>> [color_map(x) for x in range(10)]
['#e41a1c', '#e41a1c', '#626591', '#429781', '#609760', '#984ea3', '#e67328', '#ffbf19', '#e9d530', '#a65628']
```

Option 2: access the `colors` property directly. This is not a documented interface, so option 1 is better.

```
>>> color_map.colors
[(0.8941176470588236, 0.10196078431372549, 0.10980392156862745, 1.0), (0.4418300653594771, 0.36339869281045756, 0.5176470588235295, 1.0), (0.24444444444444446, 0.5581699346405229, 0.5777777777777778, 1.0), (0.3019607843137256, 0.6862745098039215, 0.2901960784313727, 1.0), (0.49803921568627446, 0.4326797385620916, 0.522875816993464, 1.0), (0.7307189542483661, 0.3699346405228759, 0.4261437908496731, 1.0), (0.9999999999999998, 0.4980392156862744, 4.2580318356211883e-16, 1.0), (1.0, 0.8326797385620914, 0.13333333333333333, 1.0), (0.8836601307189542, 0.7790849673202613, 0.18562091503267975, 1.0), (0.6509803921568628, 0.33725490196078434, 0.1568627450980392, 1.0)]

>>> import matplotlib
>>> matplotlib.colors.to_hex(color_map.colors[0])
'#e41a1c'

>>> [matplotlib.colors.to_hex(c) for c in color_map.colors]
['#e41a1c', '#715d84', '#3e8e93', '#4daf4a', '#7f6e85', '#ba5e6d', '#ff7f00', '#ffd422', '#e1c72f', '#a65628']
```

shankari (changes requested):

shankari:
Actually, let me try and run the output myself before I merge

tour_model_eval/viz_bin_cluster.py
corinne-hcr: Ah, there is somewhere online pointed out if index is given, the colormap will follow the index. So, no matter what I put on, as long as the index is correct, the colormap will be fine. I think it's not a big deal there.

corinne-hcr:
I mentioned there is some differences for clustering (min=0 and min=len(bins)).
You don't have to run it right now.
I will update it in a minute

tour_model_eval/viz_bin_cluster.py
shankari: As I said, I am not going to force you to fix this. But I didn't believe that you can't get a color list out of the colormap, so I did that little thought experiment.

shankari:
When I try to run the `viz_bins_clusters_above_cutoff` notebook, I get an error

```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-7-0f3d0ef3449a> in <module>
----> 1 bin_trips, bins = pipeline.remove_noise(trips, radius)

~/e-mission/e-mission-server/emission/analysis/modelling/tour_model/cluster_pipeline.py in remove_noise(data, radius)
     54     sim.bin_data()
     55     logging.debug('number of bins before filtering: %d' % len(sim.bins))
---> 56     sim.delete_bins()
     57     logging.debug('number of bins after filtering: %d' % len(sim.bins))
     58     return sim.newdata, sim.bins

~/e-mission/e-mission-server/emission/analysis/modelling/tour_model/similarity.py in delete_bins(self)
     97     def delete_bins(self):
     98         self.calc_cutoff_bins()
---> 99         for i in range(len(self.bins) - num):
    100             self.bins.pop()
    101         newdata = []

NameError: name 'num' is not defined
```

It looks like https://github.com/e-mission/e-mission-server/pull/792 introduced a regression in the cluster pipeline.

shankari:
Ah looks like it is fixed in https://github.com/e-mission/e-mission-server/pull/794/files
Thanks @corinne-hcr

shankari:
I can in fact run the notebook through.
Here's the graph that I get

image

tour_model_eval/viz_bin_cluster.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Plot all clusters with different colors on the map
+    # Choose the first start point to locate the map
+
+    # The function takes four parameters:
+    # - labels: trip label list after running featurization/pipeline.cluster
+    # - sel_label_list: the list of selected label (set to None for plotting all clusters)
+    # - points: coordinate points from featurization/pipeline.cluster
+    # - clusters: number of clusters
+    map = folium.Map(location=[points[0][1], points[0][0]], zoom_start=12, max_zoom=30, control_scale=True)
```
shankari, line 63: Like this

corinne-hcr:
I get this, too. I can't get the one on all data. It never stops.

tour_model_eval/viz_bin_cluster.py
```diff
@@ -0,0 +1,79 @@
...
+    # - bins: bins list above cutoff after running pipeline.remove_noise
+    # - sel_bin_indices: the list of indices of selected bins (set to None for plotting all bins)
+    # - trips: the trips after running pipeline.read_data
+
+    # In order to plot trips correctly, map index should be len(bins)+1,
+    # so that the color steps length is equal to the bins number. For bin index which is 0,
+    # will be plotted in color which is in index 1.
+    color_map = cm.linear.Set1_07.to_step(len(bins), index=[i for i in range(len(bins)+1)])
+    map = folium.Map(location=[trips[0].data.start_loc["coordinates"][1], trips[0].data.start_loc["coordinates"][0]],
+                   zoom_start=12, max_zoom=30, control_scale=True)
```
shankari, line 33: and this

shankari (changes requested):
One final fix: can you please zoom the map based on the actual coordinates instead of a hardcoded zoom. I initially thought that the comparison between the bins and clusters was not working because the trips were almost invisible at the hardcoded zoom level.

tour_model_eval/viz_bin_cluster.py
corinne-hcr: Do you mean the `zoom_start=12, max_zoom=30`? 
I don't think I can.
```zoom_start (int, default 10) – Initial zoom level for the map.```--from folium documentation
I have tried many times to set it to 12, in order to fit most of the graphs (on all data and above cutoff)

Comment (shankari) [reply tour_model_eval/viz_bin_cluster.py:33]: no, I mean that you should call `fit_bounds` to set the zoom dynamically.

**Comment by shankari** [comment]
All clusters also finished running for me

image

**Comment by shankari** [comment]
all bins | all clusters
-- | --
  | 92 | 17
 | 37 | 8
| 160 | 47
| 36 | 17
| 206 | 2
| 76 | 8
| 97 | 32
| 311 | 4
| 171 | 4
| 122 | 2
| 132 | 27
| 324 | 192

**Comment by corinne-hcr** [comment]
Can you screenshot the full data graph? (with the index) thanks!!!

**Comment by shankari** [comment]
so it looks like another difference between the bin and cluster code is the time taken to run them. 
If you can add timing statements (https://stackoverflow.com/questions/29280470/what-is-timeit-in-python), I can re-run the notebooks on my laptop and determine the actual difference.

**Comment by shankari** [comment]
> Can you screenshot the full data graph? (with the index) thanks!!!

isn't that just https://github.com/e-mission/e-mission-eval-private-data/pull/16#issuecomment-778003311

**Comment by corinne-hcr** [comment]
I think it should look like
image

**Comment by shankari** [comment]
that is not full data, but only above cutoff, right?

**Comment by corinne-hcr** [comment]
Yes, I can only get the cutoff on my laptop. But for all data, I should have set a similar data frame. Did you see the similar thing on all data? Should be a complete data frame with index(user name)

**Comment by shankari** [comment]
| above cutoff bins | above cutoff clusters (min = len(bins)) | above cutoff clusters (min = 0)
-- | -- | -- | --
user1 | 17 | 17 | 17
user2 | 9 | 9 | 9
user3 | 16 | 16 | 15
user4 | 2 | 2 | 2
user5| 20 | 20 | 19
user6| 6 | 6 | 3
user7| 7 | 7 | 7
user8| 28 | 28 | 26
user9| 27 | 27 | 26
user10 | 25 | 25 | 24
user11 | 14 | 16 | 8
user13 | 14 | 14 | 14

**Comment by shankari** [comment]
> Did you see the similar thing on all data? Should be a complete data frame with index(user name)

The dataframe for all data is at https://github.com/e-mission/e-mission-eval-private-data/pull/16#issuecomment-778005050
It is true that the index didn't get copy-pasted, but it is the same order as the cutoff - user1 ... user13 with user12 missing

**Comment by shankari** [comment]
This is better than a screenshot because you can copy-paste it and get a dataframe back without having to type out all the numbers :)

```
{'all bins': {'user1': 92,
  'user2': 37,
  'user3': 160,
  'user4': 36,
  'user5': 206,
  'user6': 76,
  'user7': 97,
  'user8': 311,
  'user9': 171,
  'user10': 122,
  'user11': 132,
  'user13': 324},
 'all clusters': {'user1': 17,
  'user2': 8,
  'user3': 47,
  'user4': 17,
  'user5': 2,
  'user6': 8,
  'user7': 32,
  'user8': 4,
  'user9': 4,
  'user10': 2,
  'user11': 27,
  'user13': 192}}
```

**Comment by corinne-hcr** [comment]
OK thanks!

**Comment by shankari** [comment]
Once you address the `fit_bounds`, I am ready to approve and merge.

File: tour_model_eval/run_time_all_data.ipynb [hunk 34]
```diff
@@ -0,0 +1,129 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "for i in range (len(all_users)):\n",
+    "    if i == 11:\n",
+    "        continue\n",
+    "    user = all_users[i]\n",
+    "    trips = pipeline.read_data(uuid=user)\n",
+    "    sim = similarity.similarity(trips, radius)\n",
+    "    %timeit sim.bin_data()"
```
shankari, line 76: where are you saving this output for combing with other results later?

shankari (changes requested):

tour_model_eval/run_time_all_data.ipynb
corinne-hcr: no, I didn't. Can't we just copy and paste the result? Or should I calculate the mean runtime?

shankari: we should not copy-and-paste the result. We should save the result to a machine-readable file. I thought we discussed this today - I distinctly remember bringing up `to_json`, `to_csv` etc.

shankari: you could also use timeit from python instead if ipython if that is easier to use, and convert the new files to `.py` files

shankari: @corinne-hcr can you maybe remove the timeit notebooks from this PR so I can merge it and move on to helping you with https://github.com/e-mission/e-mission-eval-private-data/pull/18

shankari (approved):

shankari:
@corinne-hcr merged. I squashed before merging to avoid lots of back and forth on the code redesign

PR #18, opened 2021-02-12
cannot extract user_input, need help

corinne-hcr:
cannot extract user_input after running `user_input = sim.data[i].data["user_input"]`, need help


corinne-hcr:
Screen Shot 2021-02-13 at 11 20 33 AM
I don't understand what replaced mode represents. Should I put it in comparisons as well?

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+# precision_all_users takes four parameters
+# - all_users: pass in all participants' data
+# - sp2en: default None, no need to change language
+# - cvt_purpose: default None, no need to convert purpose
+def precision_bin_all_users(all_users,radius,sp2en=None,cvt_purpose=None):
+    all_users_preci = []
+    for i in range(len(all_users)):
+        if i == 11:  # filter out non-typical user
```
shankari, line 63: In general, it is better if your filtering is based on the characteristics of the data and not the index of the user. That makes it easier to adapt to other datasets. Think of what this would have been like if Naomi had specific checks in her code from the unlabeled dataset that she used.

What makes user 11 atypical? Use that characteristic instead.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -0,0 +1,82 @@
...
+            continue
+        user = all_users[i]
+        trips = pipeline.read_data(uuid=user, key='confirmed_trip')
```
shankari, line 66: use `esda.CONFIRMED_TRIP` instead (see review of server PR)

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -0,0 +1,82 @@
...
+            if sim.data:
+                sim.bin_data()
+                if sp2en is None and cvt_purpose is None:
+                    all_bins_preci = precision_bins(all_bins_preci, sim.bins, non_empty_trips)
+                elif sp2en == 'True' and cvt_purpose is None:
+                    all_bins_preci = precision_bins(all_bins_preci, sim.bins, non_empty_trips,
+                                                    sp2en='True')
+                elif sp2en == 'True' and cvt_purpose == 'True':
+                    all_bins_preci = precision_bins(all_bins_preci, sim.bins, non_empty_trips,
+                                                    sp2en='True', cvt_purpose='True')
```
shankari, line 80:
```suggestion
                all_bins_preci = precision_bins(all_bins_preci, sim.bins, non_empty_trips, sp2en, cvt_purpose)
```

shankari (changes requested):
the `confirmed_trips_eval_all_bins` file includes outputs, so it is too large to view in GitHub. As we discussed earlier, please clear all outputs before checking in notebooks in order to avoid leaking information.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
corinne-hcr: If you don't pass in `sp2en='True'`, `cvt_purpose='True'`, can you show me how to change 
`def precision_bins(all_bins_preci,bins,non_empty_trips,sp2en=None,cvt_purpose=None):`
and how to change the way to call this function in notebook?
`all_users_preci_cvtpur=evaluation.precision_bin_all_users(all_users,radius,sp2en='True',cvt_purpose='True')`
The way I set `sp2en` and`cvt_purpose`needs to pass in `sp2en='True'`,`cvt_purpose='True'` in `precision_bin_all_users` to call `sp2en='True'` and `cvt_purpose='True'`in `precision_bins`
I change the `.py`like you did, but doesn't change the notebook, that doesn't return a correct value in the notebook, all precisions become `nan`

tour_model_eval/confirmed_trips_eval_bins_clusters.py
shankari: glad you removed this, but I don't see it replaced by a characteristic instead. What made user 11 atypical and where are you checking for that condition now?

shankari:
@corinne-hcr one more question. Would also be good if you could put the results in here just for my own edification/interest!

corinne-hcr:
What result? Reading the confirmed trips or precision scores ?

shankari:
precision scores with and without mapping. You won't be able to get them unless the confirmed trips work anyway :)

corinne-hcr:
image

shankari:
@corinne-hcr can you answer the pending question and mark this as ready to merge?

corinne-hcr:
Pending question? Do you mean the characteristic of user 11? I already delete the original filter. Since there is no trip after removing trips that are points, so there is no user_input, this user will be filtered out when I choose trips that contain user_input.

shankari (approved):

PR #19, opened 2021-02-21
v-measure

corinne-hcr:
My question for v-measure was what to put in `labels_true` and `labels_pred`. (But I have some clues now)
`metrics.homogeneity_score(labels_true, labels_pred)`
`metrics.completeness_score(labels_true, labels_pred)`
`metrics.v_measure_score(labels_true, labels_pred)`

`plot_kmeans_digits` and `plot_document_clustering` are two examples from sklearn. I add some test code to print some process.

In `plot_document_clustering`, the author uses k-means for clustering, but the data come with labels, so he uses the length of labels to determine the clusters.
```
labels = dataset.target
true_k = np.unique(labels).shape[0]
```
Then he calls KMeans like this 
```
km = KMeans(n_clusters=true_k, init='k-means++', max_iter=100, n_init=1, verbose=opts.verbose)
```
He runs v-measure like this
```
print("V-measure: %0.3f" % metrics.v_measure_score(labels, km.labels_))
```
`labels`is a 1D ndarray, and so is `km.labels_` , the size of the array is the size of all data ( so I guess I need to pass in all data labels, instead of data labels from one bin a time then calculate the mean score)
```
The type of labels is <class 'numpy.ndarray'>
the shape of labels is n_samples 3387
```
```
The type of km.labels_ is <class 'numpy.ndarray'>
the shape of km.labels_ is n_samples 3387
true_k is 4
```
```
np.unique(labels) is [0 1 2 3]
np.unique(labels).shape is 4
```
(However, in documentation examples, the author passes in lists for both `labels_true` and `labels_pred`)

Since the author didn't set `random_state` in KMeans, the `Homogeneity`,`Completeness`, and`V-measure` are different every time ( the labels from clustering are different every time)

In the documentation, there is a sentence - `This metric is independent of the absolute values of the labels: a permutation of the class or cluster label values won’t change the score value in any way.`
I understand what this means now. 
For example,
```
labels_true = [0,0,1,1]
labels_pred = [1,1,2,3]
>>>metrics.homogeneity_score(labels_true, labels_pred)
0.9999999999999999
>>>metrics.completeness_score(labels_true, labels_pred)
0.6666666666666666
>>>metrics.v_measure_score(labels_true, labels_pred)
0.7999999999999999
```
```
labels_true = [11,11,1,1]
labels_pred = [2,2,3,4]
>>>metrics.homogeneity_score(labels_true, labels_pred)
0.9999999999999999
>>>metrics.completeness_score(labels_true, labels_pred)
0.6666666666666666
>>>metrics.v_measure_score(labels_true, labels_pred)
0.7999999999999999
```
The contents above are what my see from the example codes.

Here is my questions:
For `homogeneity_score`
In documentation page, there are some examples of `homogeneity_score`, but I don't understand.
```
Clusters that include samples from different classes do not make for an homogeneous labeling:
>>> print("%.6f" % homogeneity_score([0, 0, 1, 1], [0, 1, 0, 1]))
0.0...
>>> print("%.6f" % homogeneity_score([0, 0, 1, 1], [0, 0, 0, 0]))
0.0...
```
I try a similar test on notebook.
```
labels_true = [0,1,0,1]
labels_pred = [0,0,0,0]

>>>metrics.homogeneity_score(labels_true, labels_pred)
1.6017132519074588e-16
```

The `homogeneity_score` should be bounded [0,1]. But if I run on my data, this circumstance won't happen since there are more than 1 bin/cluster. I just don't understand why would this happen.

Now, I am working on labeling the ground truth based on the user_input (on data above cutoff)--`labels_true`. I will give the same label for same user_input trips. Also, I will label the trips in bins according to the bin index--`labels_pred`. Finally, put them in the function.


shankari:
> but the data come with labels, so he uses the length of labels to determine the clusters.

Your data also comes with labels. The labels are represented by a 3 tuple of values. Why can't you use a similar technique to identify the set of unique labels?

> Now, I am working on labeling the ground truth based on the user_input (on data above cutoff)

How are you doing this?

> so I guess I need to pass in all data labels, instead of data labels from one bin a time then calculate the mean score

Yes. That is the way that all ML algorithms typically work. I think this is why we have been talking at cross-purposes for a few days.

> The homogeneity_score should be bounded [0,1]. But if I run on my data, this circumstance won't happen since there are more than 1 bin/cluster. I just don't understand why would this happen.

I don't understand this statement. The score in the notebook is `1.6017132519074588e-16` which is between [0,1] and is 0 if printed in the same way as the example.

```
>>> print("%.6f" % 1.6017132519074588e-16)
0.000000
```

corinne-hcr:
Thanks for clarification on `1.6017132519074588e-16`
> (e.g. drop_duplicates)

Yes, that's what I am doing

corinne-hcr:
@shankari I have a question about user_input
What's the difference between 
51 | pilot_ebike | home | pilot_ebike
-- | -- | -- | --

 and 
98 | pilot_ebike | home | same_mode
-- | -- | -- | --

??

If I remember correctly, `same_mode` in `replaced_mode` is the same as the one in `mode_confirm`

shankari:
correct. There is no difference. Both of those are invalid `replaced_mode` for pilot eBike since we wanted to know what mode people would have used if they didn't have the eBike, but not everybody fixed their labels. The mode and purpose are valid, though.

shankari:
There should be very few entries like that

corinne-hcr:
I haven't counted all users.
For user[0],
```
['pilot_ebike', 'community_walk', 'same_mode']
['pilot_ebike', 'home', 'same_mode']
['pilot_ebike', 'home', 'same_mode']
['walk', 'home', 'same_mode']
```
For user[2]
```
['pilot_ebike', 'work', 'same_mode']
['pilot_ebike', 'work', 'same_mode']
['pilot_ebike', 'work', 'same_mode']
['pilot_ebike', 'shopping', 'same_mode']
['pilot_ebike', 'work', 'same_mode']
['drove_alone', 'work', 'same_mode']
['drove_alone', 'work', 'same_mode']
['drove_alone', 'pick_drop', 'same_mode']
['drove_alone', 'pick_drop', 'same_mode']
```
For user[3]
```
['walk', 'pick_drop', 'same_mode']
['walk', 'home', 'same_mode']
```
For user[4], there are some NaN in both `purpose_confirm` and `replaced_mode`, the score is lower.
```
  mode_confirm purpose_confirm replaced_mode
0  shared_ride             NaN           NaN
1  drove_alone        shopping           NaN
2  drove_alone            work     no_travel
3  shared_ride            work           NaN
  mode_confirm purpose_confirm replaced_mode
0  pilot_ebike   entertainment   drove_alone
1  drove_alone             NaN           NaN
2  pilot_ebike            home   drove_alone
  mode_confirm purpose_confirm replaced_mode
0  pilot_ebike   entertainment   drove_alone
1  drove_alone             NaN           NaN
2  pilot_ebike            home   drove_alone
```

corinne-hcr:
At this point, I am going to get v-measure for original user_input, after changing language, and converting purposes and mode, on all bins, bin above cutoff, and clusters above cutoff

@shankari suggested 

> we should then move on to the number of user requests per month as the other part of the tradeoff 
>     and to the second task to see if we can bump up the accuracies. As part of that, I really want to try out other clustering     techniques as well such as DBSCAN
  

>  as you are working on the binning, I would suggest making a wrapper around the bin algorithm that is consistent with the  sklearn interfaces
> ​
>     that will make it easy for you to experiment with multiple algorithms using sklearn (https://scikit-learn.org/stable/modules/model_evaluation.html)
>     e.g. https://scikit-learn.org/stable/modules/grid_search.html#grid-search
> 

> sklearn has a specific interface that it expects (e.g. https://scikit-learn.org/stable/developers/develop.html). If you implement that interface, you can use the built-in GridSearch (https://scikit-learn.org/stable/modules/grid_search.html#exhaustive-grid-search) without having to call the functions directly

> one caveat is that I am not sure whether GridSearchCV works well for clustering (as opposed to classification) algorithms (https://stackoverflow.com/questions/25633383/how-can-gridsearchcv-be-used-for-clustering-meanshift-or-dbscan)

shankari:
> For user[4], there are some NaN in both purpose_confirm and replaced_mode, the score is lower.

I don't recall seeing any `NaN` in the purpose entries.

I double-checked the entries using 
https://github.com/e-mission/em-public-dashboard/blob/main/viz_scripts/mode_purpose_share.ipynb

I suspect this may be because of your mapping code. But you can try the code from the public dashboard to see if you still have `N/A` entries and if not, debug the differences between that code and yours.

```
>>> expanded_ct["purpose_confirm"].unique()

array(['meal', 'personal_med', 'transit_transfer', 'shopping', 'home',
       'exercise', 'work', 'entertainment', 'pick_drop', 'school',
       'revisado_bike', 'placas_de carro', 'on_the way home',
       'work_travel', 'no_travel', 'work_break - short walk',
       'work_- lunch break',
       'friend_was running errands before dropping me off after work',
       'multiple_errands, etc.', 'lunch_break', 'break', nan, 'religious',
       'bank', 'iglesia', 'community_walk', 'gardening', 'visiting',
       'church', 'community_garden', 'community_meeting',
       'visit_a friend', 'aseguranza', 'gas', 'mi_hija recién aliviada',
       'servicio_comunitario', 'pago_de aseguranza',
       'caminata_comunitaria'], dtype=object)
```

shankari:
> For user[4], there are some NaN in both purpose_confirm and replaced_mode, the score is lower.

I don't recall seeing any `NaN` in the purpose entries.

I double-checked the entries using 
https://github.com/e-mission/em-public-dashboard/blob/main/viz_scripts/mode_purpose_share.ipynb

I suspect this may be because of your mapping code. But you can try the code from the public dashboard to see if you still have `N/A` entries and if not, debug the differences between that code and yours.

```
>>> expanded_ct["purpose_confirm"].unique()

array(['meal', 'personal_med', 'transit_transfer', 'shopping', 'home',
       'exercise', 'work', 'entertainment', 'pick_drop', 'school',
       ...], dtype=object)
```

corinne-hcr:
I didn't convert anything from this user, the output is just after filtering out empty user_input from non_empty_trips. 
For this user, total 453 trips, but only 33 trips have user_input, after filtering out trips that are points, 28 left.
```

mode_confirm | purpose_confirm | replaced_mode

pilot_ebike | entertainment | drove_alone
pilot_ebike | entertainment | drove_alone
shared_ride | NaN | NaN
drove_alone | NaN | NaN
drove_alone | NaN | NaN
pilot_ebike | NaN | NaN
drove_alone | NaN | NaN
shared_ride | NaN | NaN
pilot_ebike | NaN | NaN
drove_alone | NaN | NaN
drove_alone | NaN | NaN
pilot_ebike | exercise | NaN
drove_alone | NaN | NaN
drove_alone | work | no_travel
drove_alone | NaN | no_travel
drove_alone | shopping | NaN
drove_alone | exercise | no_travel
drove_alone | shopping | NaN
drove_alone | work | no_travel
drove_alone | exercise | no_travel
drove_alone | NaN | NaN
drove_alone | NaN | NaN
pilot_ebike | home | drove_alone
pilot_ebike | home | drove_alone
shared_ride | work | NaN
drove_alone | entertainment | NaN
drove_alone | work | NaN
drove_alone | work | no_travel
```

corinne-hcr:
So, should I map `same_mode` to the mode in `mode_confirm`?

shankari:
@corinne-hcr I am not sure which user you are talking about; the order that you and I have may not be the same. I don't think we should share UUIDs in the public forum, and by design, the public dashboard only performs aggregate analysis.

But I can confirm that I don't see any `N/A` entries for `purpose_confirm` across all participants. Are you sure you are filtering the users to only look at participants?

shankari (changes requested):

shankari:
> So, should I map same_mode to the mode in mode_confirm?

Sure

shankari:
> Do you really have a different output from mine?

> But I can confirm that I don't see any N/A entries for purpose_confirm across all participants.

Did you investigate and determine the reason for the discrepancy?

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
```diff
@@ -0,0 +1,271 @@
...
+   "id": "framed-sector",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# collect labels_pred based on bins\n",
+    "labels_pred = []\n",
+    "for i in range(len(bins)):\n",
+    "    for trip in bins[i]:\n",
+    "        labels_pred.append(i)\n",
+    "labels_pred"
```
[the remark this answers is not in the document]
corinne-hcr: I think they are in same order.

You can see the `bin_df`
Screen Shot 2021-02-21 at 11 17 43 AM
That's the trips order after binning. 
It is according to the `non_empty_trips`indices in the bins
image

The following is the list from where I collect `labels_true`, it is indeed according to the indices in `bin_trips`
image

I do consider the trip order. I think if `bin_trips_user_input_ls` has the same trip order as trips in `bin_df`, then the way I collect labels should be the same. Although I just show part of the trips, but you can see from the picture, the trips are in same order.

Is that clear?

corinne-hcr:
The following is my investigation.
I copy your `scaffolding`code, then choose the specific uuid which I want to look at

In my file, the trips are from 2020/11 to 2021/1
So, I screenshot those three months labeled trips from your`mode_purpose_share` to compare

2020/11
image
2020/12
image
2021/1
image

Here is` non_empty_trips` after binning from my file
```
  mode_confirm purpose_confirm replaced_mode
0  shared_ride             NaN           NaN
1  drove_alone        shopping           NaN
2  drove_alone            work     no_travel
3  shared_ride            work           NaN
  mode_confirm purpose_confirm replaced_mode
0  pilot_ebike   entertainment   drove_alone
1  drove_alone             NaN           NaN
2  pilot_ebike            home   drove_alone
  mode_confirm purpose_confirm replaced_mode
0  pilot_ebike   entertainment   drove_alone
1  drove_alone             NaN           NaN
2  pilot_ebike            home   drove_alone
```

So, you can see, the total labeled trips from your code are the same from my code. I also got NaN output from your code. I think that's because the user didn't input value.

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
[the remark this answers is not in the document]
shankari: I still don't see this.

`bin_user_input` that you have listed above uses `bins` not `bin_trips`, the same as `labels_pred`.
I think that your main argument is that the labels are the same through spot checking. That is good, but it could also happen by chance and for the first few entries. I certainly did not look at every single entry, and I am not sure I would be very accurate at doing so. This is the kind of checking which it is much better for computers to do programatically.

I would feel a lot more confident by comparing the trip indices in the original list directly. Add an index to the original list (if it doesn't have one) and use it as the dataframe index for both `labels_pred` and `labels_true` and verify that the indices are the same.

shankari:
can you put the related UUID into the teams chat? As I said earlier, I don't see `NaN` as a value in purpose_confirm when I run it

```
expanded_ct["purpose_confirm"].unique()
```

does not include `NaN`

shankari:
```
>>> expanded_ct["purpose_confirm"].isnull()
106     False
108     False
109     False
111     False
113     False
        ...  
4105    False
4106    False
4107    False
4108    False
4109    False
Name: purpose_confirm, Length: 2425, dtype: bool
```

corinne-hcr:
I think you should pick 2020/12 or 2021/1

shankari:
I'm not sure what you mean by that. I still need to know the user so that the indices match up, right?

corinne-hcr:
I have already put the uuid in Teams. In 2020/11, there is no NaN for this user, but in 2020/12 and 2021/1, you can see NaN

shankari:
nvm, you are right!

Instead of manually looking at the `isnull()` output, I checked it using code, and there are some missing entries, though they don't match your indices above.

```
>>> pc = expanded_ct["purpose_confirm"].isnull()
>>> pc[pc == True]
1919    True
2095    True
2129    True
2131    True
2132    True
2136    True
2800    True
2805    True
2807    True
2809    True
2811    True
2813    True
2914    True
2916    True
2918    True
2934    True
2940    True
2945    True
3039    True
3041    True
3899    True
3909    True
```

shankari:
And double checking those entries from the dataframe does indicate that they are missing.

```
pc = expanded_ct["purpose_confirm"].isnull()
expanded_ct.loc[pc[pc == True].index][["mode_confirm", "purpose_confirm", "replaced_mode"]]
```


mode_confirm | purpose_confirm | replaced_mode
 -- | -- | --
pilot_ebike | NaN | NaN
pilot_ebike | NaN | NaN
pilot_ebike | NaN | NaN
pilot_ebike | NaN | NaN
pilot_ebike | NaN | NaN
pilot_ebike | NaN | taxi
shared_ride | NaN | NaN
drove_alone | NaN | NaN
drove_alone | NaN | NaN
pilot_ebike | NaN | NaN
drove_alone | NaN | NaN
shared_ride | NaN | NaN
pilot_ebike | NaN | NaN
drove_alone | NaN | NaN
drove_alone | NaN | NaN
drove_alone | NaN | NaN
drove_alone | NaN | NaN
drove_alone | NaN | no_travel
drove_alone | NaN | NaN
drove_alone | NaN | NaN
not_a_trip | NaN | no_travel
not_a_trip | NaN | no_travel

shankari:
I think we should skip these entries for now. There are not many of them, and the user has only a few labeled trips to begin with. So I wasn't counting him in my mental training set anyway.

You can just skip all users with < 50% labeling rate for this training and tuning stage.

corinne-hcr:
Btw, this is the way how Naomi collected `bin_trips` above cutoff bins
```
    #delete lower portion of bins
    def delete_bins(self):
        self.calc_cutoff_bins()
        for i in range(len(self.bins) - num):
            self.bins.pop()
        newdata = []
        for bin in self.bins:
            for b in bin:
                d = self.data[b]
                newdata.append(self.data[b])
        self.newdata = newdata if len(newdata) > 1 else self.data
```
`bin_trips` is `self.newdata`, and it is exactly according to the indices from bins.
And the indices in bins

Here is how the she collected indices in bins
```
        logging.debug('After removing trips that are points, there are %s data points' % len(self.data))
        self.size = len(self.data)

    #create bins
    def bin_data(self):
        for a in range(self.size):
            added = False
            for bin in self.bins:
                try:
                    if self.match(a,bin):
                        bin.append(a)
                        added = True
                        break
                except:
                    added = False
            if not added:
                self.bins.append([a])
        self.bins.sort(key=lambda bin: len(bin), reverse=True)
```
Here `a` is the original index of the trip after she filtered out trips that are points.
So, in algorithm, the trips order in `bins` is exactly the same as trips order in `bin_trips`

corinne-hcr:
> You can just skip all users with < 50% labeling rate for this training and tuning stage.

Do you mean among all users labeled trips, I should skip users that have <50% full labeled trips in the three columns ?

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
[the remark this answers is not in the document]
corinne-hcr: The red list of `bins` above contains the original indices from trips after Naomi filtered out trips that are points. `bin_trips` is a new data list that collects trips according to the indices in bins. 
Which key word is for timestamp that you want me to add?

shankari:
> Do you mean among all users labeled trips, I should skip users that have <50% full labeled trips in the three columns ?

Correct. You will end up with ~ 6 users I think.

shankari:
We can add 4 other users from another project later; their confirmed labels will be a 2-tuple of `mode_confirm` and `purpose_confirm` only. So ~ 10 users but for multiple months each

corinne-hcr:
Can we just use what we have now for further exploration? Because if the user input format or completeness is different, all work I have done including what I am doing now have to be modified. It is not quite worthy to spend time on repeating work.
For now, I need to filter out the users who don't have 50% of full(3) labeled user inputs. If the later 4 users only have 2 kinds of labels, they don't match the algorithm. If I change the metric to fit that 4 users, the other 6 users may contain NaN. It is not efficient to change the same thing again and again.

shankari:
> Because if the user input format or completeness is different, all work I have done including what I am doing now have to be modified. 

We should be able to write code that is not so specific to the exact dataset that you are working with.

Again, in terms of generalization, results on a dataset with only 6 users are going to be less impressive than results on a dataset with 10 users (almost twice as many!)

I understand that you are not used to writing generalizable code (as shown from the plotting example that you wrote earlier). So submit updates to the PR often and I will show you how to make it generalizable.

shankari:
For example, this will generate non-empty entries in the dataframe in a way that it is super easy to change later.

```
USER_INPUT_VALUES = ["mode_confirm", "replaced_mode", "purpose_confirm"]
pc = expanded_ct[USER_INPUT_VALUES].dropna()
expanded_ct.loc[pc[pc == True].index][USER_INPUT_VALUES]
```

shankari:
And finally, if it is too hard for you to filter out the entries with `NaN`, just retain them. They are all from users with < 50% confirmed trips (`len(user_input) > 0`) anyway so they won't make a huge difference.

It is vastly more important to have results from 10 users than to be picky about the 22 trips which have a NaN in one field.

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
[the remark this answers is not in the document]
shankari: `trip["data"]["start_ts"]` should work as a timestamp. Again, I don't want you to print it, I want you to check it programmatically

corinne-hcr:
If later the 4 users only have 2 kinds of user_input, should we just based on those 2 kinds of user_input to filter out NaN trips ( which have NaN values on those 2 columns). And all comparisons are only based on those 2 columns instead of 3 columns? 
All comparisons now are based on 3 columns.

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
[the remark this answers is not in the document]
corinne-hcr: image
Is that ok?

shankari:
@corinne-hcr What I'm saying in https://github.com/e-mission/e-mission-eval-private-data/pull/19#issuecomment-783081804 is that if it is going to take you a long time to filter NaN trips, and you don't know how to do it generically, don't filter them.

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
```diff
@@ -0,0 +1,271 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "tired-athens",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# turn all user_input into list without binning\n",
+    "bin_trips_user_input_ls = pd.DataFrame(data=[bin_trips[i][\"data\"][\"user_input\"] for i in range(len(bin_trips))]).values.tolist()\n",
```
shankari, line 149: can you avoid index based loops, which are not pythonic? For example, this could be:

shankari, line 149:
```suggestion
    "bin_trips_user_input_ls = pd.DataFrame(data=[bt[\"data\"][\"user_input\"] for bt in bin_trips]).values.tolist()\n",
```

shankari, line 149: Also, why are you switching back and forth between dataframes and lists? Why isn't this just:

shankari, line 149:
```suggestion
    "bin_trips_df = pd.DataFrame(data=[bt[\"data\"][\"user_input\"] for bt in bin_trips])\n",
```

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
```diff
@@ -0,0 +1,271 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "veterinary-courage",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "for i in bin_trips_user_input_ls:\n",
+    "    if i[2] == \"same_mode\":\n",
+    "        print(i)"
+   ]
```
shankari, line 163: And this is:

```suggestion
    "bin_trips_df[bin_trips_df.replaced_mode == \"same_mode\"]:\n",
   ]
```

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
```diff
@@ -0,0 +1,271 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "coastal-fiction",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# drop duplicate user_input\n",
+    "no_dup_df=pd.DataFrame(data=[bin_trips[i][\"data\"][\"user_input\"] for i in range(len(bin_trips))]).drop_duplicates()\n",
+    "no_dup_df"
+   ]
```
shankari, line 175: Then this is just:
```suggestion
   "source": [
    "# drop duplicate user_input\n",
    "no_dup_df= bin_trips_df.drop_duplicates()\n",
    "no_dup_df"
   ]
```

Switching back and forth between pandas dataframes and lists is inefficient - the dataframes are designed to be efficient for multiple sequential operations.

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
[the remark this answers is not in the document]
shankari: That looks good, but not using the full power of pandas/numpy. Please add to PR.

shankari (changes requested):

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
corinne-hcr: The list is used to compare with the non duplicate list. I think there should be a way to find out if a certain row in one data frame is in another data frame, then find out the index of that row in another data frame. But I didn't figure out how to do that.

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
```diff
@@ -3,7 +3,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "introductory-sunglasses",
+   "id": "severe-married",
```
shankari, line 6: I'm going to stop commenting on id changes after this :)

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
```diff
@@ -216,10 +230,55 @@
...
+  },
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "unlimited-waste",
+   "id": "dental-universal",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "bin_trips_ts = pd.DataFrame(data=[i[\"data\"][\"start_ts\"]for i in bin_trips])\n",
```
shankari, line 248: Readability improvement: `i` is generally an index (e.g. 2) `trip` is a trip object that you can get data from

```suggestion
    "bin_trips_ts = pd.DataFrame(data=[trip[\"data\"][\"start_ts\"]for trip in bin_trips])\n",
```

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
```diff
@@ -138,47 +141,58 @@
...
+   "id": "historic-russian",
    "metadata": {},
    "outputs": [],
    "source": [
-    "for i in bin_trips_user_input_ls:\n",
-    "    if i[2] == \"same_mode\":\n",
-    "        print(i)"
+    "# turn all user_input into list without binning\n",
+    "bin_trips_user_input_ls = bin_trips_df.values.tolist()\n",
+    "bin_trips_user_input_ls"
```
shankari, line 172: where are you using this? can you just delete it?

shankari (changes requested):
Very close, just some small changes needed.

tour_model_eval/v-measurel_cutoff_bins_single_user.ipynb
corinne-hcr: I use `bin_trips_user_input_ls` to get `labels_true` because I don't know how to directly check if one row in a data frame is also in another data frame and get the index from another data frame.

corinne-hcr:
This is v-score for all bins
image


This is v-score for bins above cutoff
image

This is v-score for above cutoff cluster based on sil score
image

Take user[6] as example, on all bins, non-duplicate trips are 18 (without any changes), 42bins; on bins above cutoff, non-duplicate trips are 8 (without any changes), 7 bins; on clusters above cutoff based on sil score, non-duplicate trips are 8 (without any changes), 5 clusters. So, the score is obviously different. When the number of bins/cluster is greater than non-duplicate trips number, the score will go up. We need cut bins/cluster above cutoff into smaller bins/clusters.

But I haven't changed precision code and the precision on all bins. Since we change the bar to evaluate valid users, the previous precision code doesn't work well for the new bar. Should I change them or delete them?

shankari:
> But I haven't changed precision code and the precision on all bins. Since we change the bar to evaluate valid users, the previous precision code doesn't work well for the new bar. Should I change them or delete them?

It is fine to delete.

> When the number of bins/cluster is greater than non-duplicate trips number, the score will go up. We need cut bins/cluster above cutoff into smaller bins/clusters.

First, you need to put this into a graph (when I saw that there was a picture, I thought you had already made it into a graph).
Second, the homogeneity scores for the above cutoff are actually pretty bad; there are some that are actually below 0.5.
The obvious next step is to tune the threshold; narrowing it down to 100 meters to see if it helps.
But I think that a better option is actually to go ahead and write a notebook to compute the other part of the tradeoff (the number of times we will prompt the users per month) and then move on to stage 2.

Then we can explore multiple options for improving the score - narrowing the threshold, including time of day, including distance and duration,... and evaluate the tradeoff for each of them.

shankari:
Because my gut feeling is that narrowing the threshold is not going to be enough

shankari:
can you clean up this PR and mark it ready for review so I can merge it?

corinne-hcr:
I didn't add additional code for 2 user_input data since I haven't seen the data, I don't know how to make them match. It is only for 3 user_input now.

shankari:
that's fine. I want to be checking in code as we go along.

corinne-hcr:
This is query time for all users
image

radius = 100, and I filter our users that have 0 valid trips instead of valid trips < 10

shankari:
@corinne-hcr make sure to add the number of bins above the cutoff as well, since we will need to query the users at least once for them :) Otherwise we won't have the labels that we can use for the subsequent trips!

shankari:
Also, are these the means? The max? Because 3 * 30 is not 22

corinne-hcr:
> Specially for this, mean is NOT going to capture the distribution.

I have question about this. The way I calculate the query times for a day is
`query_day.append(math.ceil(mean(query_day_ls)))`
See the code, I bin the trips on the same day, then put the length of each bin in `query_day_ls`, then calculate the mean like this way to get `query times in a day`, is that correct? I don't understand that mean is NOT going to capture the distribution.

shankari:
> I don't understand that mean is NOT going to capture the distribution.

Mean is only one statistical measure of a distribution, and is not always the most appropriate one.
Particularly for this distribution, which is likely to be skewed with lots of queries at the beginning and (hopefully) few later.

Search for "mean for skewed distribution" to learn more. That is also why I want you to generate graphs for the scores as well instead of relying on mean.

corinne-hcr:
I have question about _ plot the values and generate graphs_.
Let's 30 days in a month, but the data from below cutoff are not from all 30 days, maybe just several days in a month. So, when I calculate the query times for a day, I bin the trips that in a same day. Then there would be 3 times/trips in a day, 4 times, 5....and then ceil the mean. 
Then the problem is, I have 13 users, 11 of them have data. From those 11 users, every one of them have day query times and month query times. If I am not using mean, should I plot 11graphs for day, 11 graphs for month? Mean, Median, Mode, which one should I use?

Same question to v-score graph. I have three situations (original input, after changing languages, and after converting purposes and modes), 13 users, three kinds of scores (homo, completeness, v-score). Should I plot a graph/frame for homo on 13 users, completeness for 13 users, and v-score for 13 users, instead of putting them in one frame?

shankari:
@corinne-hcr yes, you would plot 11 different curves, either on one plot or on multiple subplots of one plot.

corinne-hcr:
> So you will need one entry for each _bin_ above the cutoff, and one entry for each _trip_ below the cutoff.

I am not sure if I understand it correctly. For each bin above cutoff, it should represent a type of trip. So, the length of bins above cutoff represents the types of trips (we only ask once for each type of trip). For trips below cutoff, we need to ask the users for each of them. Therefore, the total number of queries would be the number(length) of bins above cutoff and the number of trips above cutoff. As counting the day_query or month_query for bins above cutoff, I need find the earliest day/month of a trip in one bin, thus after the first query, the rest of the trips in that bin would not be asked again. Am I right?

For 11 curves, should I use a line chart? Should I plot a big chart for day_query that contains 11 users and the same for month_query? To be detailed, for each subplot, the x-axis is the query times in a day, the y-axis represents the frequency(number) of a certain query times (for example, the query times in a day is[2,4,4,4,5,6], then for x=4, y=3) , correct?
And I don't need to find mean, median, or mode for query_day and query_time?

For the three scores, one big graph for homogeneity, it includes three subplots(original input, after changing languages, after converting purposes and modes) and each plot has scores for all valid users. And the same for completeness and v-score. Don't use mean in any plot. Correct?

shankari:
> Therefore, the total number of queries would be the number(length) of bins above cutoff and the number of trips below cutoff.

Right, this is what I said.

> Am I right?

Right

As for the chart details, your instincts are generally right. Show me the results and I can give you additional feedback.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,206 @@
...
-# - sp2en: default None, no need to change language
-# - cvt_purpose: default None, no need to convert purpose
-def precision_bin_all_users(all_users,radius,sp2en=None,cvt_purpose=None):
-    all_users_preci = []
+
+def filter_data(user,radius):
+    trips = pipeline.read_data(uuid=user, key=esda.CONFIRMED_TRIP_KEY)
+    non_empty_trips = [t for t in trips if t["data"]["user_input"] != {}]
+    valid_trips = [t for t in non_empty_trips if 'mode_confirm' in t["data"]["user_input"] and
+                   'purpose_confirm' in t["data"]["user_input"] and 'replaced_mode' in t["data"]["user_input"]]
```
shankari, line 30: This is combining valid trips and binning. In the next PR, I would pull out the valid trip filtering into a separate function so it can be reused across the various modules - v-score calculation, query calculation, future clustering, etc.

shankari: also, is there a reason you are not using `dropna` here instead of hardcoding the entry checks?

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,206 @@
...
-        all_users_preci.append(round(mean(all_bins_preci), 2))
-    return all_users_preci
+        filter_trips,sim = filter_data(user,radius)
+
+        # filter out users that haven't enough trips (at least 10) to analyze
+        if len(filter_trips) < 10:
+            homo_score.append(NaN)
+            comp_score.append(NaN)
+            v_score.append(NaN)
+            continue
```
shankari, line 55: Again, as we discussed, please combine this with the other validity checks.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,206 @@
...
+        elif cutoff:
+            sim.delete_bins()
+            bin_trips = sim.newdata
+            bins = sim.bins
+
+            if len(bin_trips) < 10:
+                homo_score.append(NaN)
+                comp_score.append(NaN)
+                v_score.append(NaN)
+                continue
```
shankari, line 74: Ditto

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,206 @@
...
+    for i in range(len(all_users)):
+        user = all_users[i]
+        filter_trips,sim = filter_data(user,radius)
+
+        # filter out users that haven't enough trips (at least 10) to analyze
+        if len(filter_trips) < 10:
+            homo_score.append(NaN)
+            comp_score.append(NaN)
+            v_score.append(NaN)
+            continue
```
shankari, line 144: Ditto

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,206 @@
...
+        sim.delete_bins()
+        bin_trips = sim.newdata
+        bins = sim.bins
+
+        if len(bin_trips) < 10:
+            homo_score.append(NaN)
+            comp_score.append(NaN)
+            v_score.append(NaN)
+            continue
+
```
shankari, line 155: Ditto

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,206 @@
...
+        # turn non-duplicate user_input into list
+        no_dup_list = no_dup_df.values.tolist()
+        # collect labels_true based on user_input
+        labels_true = []
+        for trip in cluster_user_input_ls:
+            if trip in no_dup_list:
+                labels_true.append(no_dup_list.index(trip))
+        labels_pred = feat.labels
+
+        # compare the points in cluster_trips and those in feat.points, return nothing if two data frames are the same
```
shankari, line 189: can you fix the comment here? Right now if the frames are the same, we just go on to the score calculation

shankari (changes requested):
Can you please make sure to remove outputs before committing? I am finding this hard to review.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,213 @@
...
+
+
+def filter_data(user,radius):
+    trips = pipeline.read_data(uuid=user, key=esda.CONFIRMED_TRIP_KEY)
+    non_empty_trips = [t for t in trips if t["data"]["user_input"] != {}]
+    non_empty_trips_df = pd.DataFrame(t["data"]["user_input"] for t in non_empty_trips)
+    valid_trips_df = non_empty_trips_df.dropna(axis=0, how='any', thresh=None, subset=None, inplace=False)
+    valid_trips_idx_ls = valid_trips_df.index.tolist()
+    valid_trips = [non_empty_trips[i]for i in valid_trips_idx_ls]
+
```
shankari, line 40: why do you keep switching to and from dataframes?

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,213 @@
...
+    sim = similarity.similarity(valid_trips, radius)
+    filter_trips = sim.data
+    return filter_trips,sim,trips
```
shankari, line 43: why do you determine invalid trips and call similarity at the same time?
it seems like they are two separate steps.
I would create a `get_valid_trips` which only returns valid trips and call `similarity` separately

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,213 @@
...
+
+def valid_user(filter_trips,trips):
+    valid = False
+    if len(filter_trips) >= 10 and len(filter_trips) / len(trips) >= 0.5:
```
shankari, line 47: if I understand your code correctly, you are determining valid users by looking at `filter_trips` which are the trips above the cutoff. Seems like you should use `valid_trips` (the trips with valid labels) instead. Not sure that will make a difference for the results, but it is also a simple fix.

shankari (commented):
What is the difference between the various notebooks here?
Can you add a description at the top of each notebook indicating what it does.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
shankari: ah I see that you have not called `bin_data` so the `filter_trips` are the same as the `valid_trips`. But then why pass it in to similarity in the first place?

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,213 @@
...
+        if sp2en:
+            bin_trips_df = bin_trips_df.replace(span_eng_dict)
+        elif cvt_pur_mo:
+            bin_trips_df = bin_trips_df.replace(span_eng_dict)
+            bin_trips_df = bin_trips_df.replace(map_pur_dict)
+            for a in range(len(bin_trips_df)):
+                if bin_trips_df.iloc[a]["replaced_mode"] == "same_mode":
+                    # to see which row will be converted
+                    logging.debug("The following rows will be changed: %s", bin_trips_df.iloc[a])
+                    bin_trips_df.iloc[a]["replaced_mode"] = bin_trips_df.iloc[a]['mode_confirm']
```
shankari, line 98: This code has been duplicated twice - seems like it should be possible to pull it out.
Ideally, you would set this up as a pipeline in which this filtering happens first before the binning/clustering

tour_model_eval/confirmed_trips_eval_bins_clusters.py
corinne-hcr: Because I need to collect valid trips in a list for further processing. The purpose of turning non_empty_trips to Dataframe is to use dropna() to get valid trips

tour_model_eval/confirmed_trips_eval_bins_clusters.py
corinne-hcr: The first step of `similarity` code is to filter out trips that are points. I pass in valid trips, it will remove points in valid trips.

shankari (changes requested):
I would split the code into multiple functions so that they can be composed into a pipeline.

the functions would be:
- get_valid_trips
- map_labels (sp2en or cvt_pur_mo)
- valid_user_check
- fit_clusters (with two implementations, bin and cluster)
- get_pred_labels (two implementations?)
- get_user_labels (two implementations?)
- verify_clusters (with two implementations)
- compute scores

Then you have two pipelines, one with the bin implementations and the other with the cluster implementations, calling one after the other without any intermediate steps or repetition.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
shankari: can you please add that as a comment then? It is not clear otherwise.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
shankari: the right long-term solution for that is to change similarity to take in a dataframe as input.
but we can defer that solution until later

tour_model_eval/confirmed_trips_eval_bins_clusters.py
corinne-hcr: `filter_trips` are just valid trips after filtering out the points, not trips above cutoff.
All trips that are used for binning and clustering are based on `filter_trips`, `trips` are original trips, so valid users come from here.

shankari:
On a second look, the current pipelines are not too bad since several of the steps will have separate implementations anyway. I will accept a simpler refactoring in which you pull out the common functionality `map_labels`, `valid_user_check` and `compute scores` into separate functions which you call from the implementations.

If you want to implement the cleaner, but more complex refactoring, I won't complain!

tour_model_eval/confirmed_trips_eval_bins_clusters.py
shankari: I saw that 

> ah I see that you have not called bin_data so the filter_trips are the same as the valid_trips. But then why pass it in to similarity in the first place?

But without a comment about the "filtering out the points" to explain what you are doing, the reason for this is unclear. Please add the comment.

corinne-hcr:
I can pull out `map_labels`, but `valid_user_check` is already a single function, it actually based on `filter_trips` and `trips`, which are from `filter_data`(another separate function). Because I need different return values for the following steps from these two functions, I don't make them into one function.

shankari:
> valid_user_check is already a single function,

`valid_user_check` is not a single function, `valid_user` is.

> Because I need different return values for the following steps from these two functions, I don't make them into one function.

Not sure what this means. The code that uses `valid_user` appears to be identical in both cases. Can you point out where it is different?

```

        if not valid_user(filter_trips,trips):
            homo_score.append(NaN)
            comp_score.append(NaN)
            v_score.append(NaN)
            continue
```

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -20,54 +22,213 @@
...
+                bin_ls.append(index)
+        bins_ts = pd.DataFrame(data=[filter_trips[i]["data"]["start_ts"] for i in bin_ls])
+        # compare two data frames, the program will continue to score calculation if two data frames are the same
+        assert_frame_equal(bins_ts, bin_trips_ts)
+        homo = metrics.homogeneity_score(labels_true, labels_pred)
+        homo_score.append(float('%.3f' % homo))
+        comp = metrics.completeness_score(labels_true, labels_pred)
+        comp_score.append(float('%.3f' % comp))
+        v = metrics.v_measure_score(labels_true, labels_pred)
+        v_score.append(float('%.3f' % v))
```
shankari, line 133: This should be pulled out into `compute_score`.
In general, you do not want to have any repeated code.
I think I have linked to DRY before 
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

corinne-hcr: OK, I understand `valid_user_check` and `compute_score` now.
Can I modify the code after I figure out the solution for the second round clustering?

shankari:
I was hoping you would finish the refactoring and add the descriptions for the other notebooks first so I could merge this.
You could then submit a new PR for the second round clustering so I review your code properly instead of being relegated to view it over screenshare. I cannot review code when somebody else is navigating it remotely.

corinne-hcr:
For the scatter, the one I have is based on the bins above cutoff. The trips are grouped only based on coordinates and radius. Back then we compared the difference between bins and clusters(min_clusters = 0) and got that the bins are better than clusters. But when we decided to use a second round clustering, I changed to use k-means for the first round clustering and set min_clusters = len(bins). So, now should I just leave like that? Later when I have result from hierarchical clustering, I write another code for the improved scatter?

shankari:
@corinne-hcr 

1. Not sure that you need to change to kmeans for the first round. You could use bins for the first round and agglomerative clustering for the second round.
2. IIRC, the results for the bins above cutoff and clusters above cutoff were not that different. So I don't think it matters a lot which you use for the first round.

not sure about your question - are you asking whether you need to go back and change the results for the trip end only clusters? In general, for any results, you don't have to present only one set of results. You can present a set of results and pick the best one to explore further.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
shankari: @corinne-hcr  I don't see this comment in the PR

tour_model_eval/confirmed_trips_eval_bins_clusters.py
shankari: @corinne-hcr I understand that you want to wait until the end to share the code. My concern is that then the code will be large and unwieldy and impossible to review. I would really like to review and merge intermediate code as we move along.

Can you please get this PR into a reviewable state ASAP so I can merge?

shankari (changes requested):
@corinne-hcr please get this PR to a reviewable state before continuing. It is very hard to review large chunks of code with no context. I'm sure that the code is clear to you, but I'd like (at least this time) for the code to be relatively clear to others as well.

That means digestible code chunks with lots of comments explaining what you are doing.
Please get this done *before* you commit your refactored code.

corinne-hcr:
I add some comments. Please let me know if it is still not digestible.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -9,65 +9,252 @@
...
+# append NaN to the score lists when the user invalid
+def valid_user_check(filter_trips,trips,homo_score,comp_score,v_score):
+    if not valid_user(filter_trips, trips):
+        homo_score.append(NaN)
+        comp_score.append(NaN)
+        v_score.append(NaN)
+        skip = True
+    else:
+        skip = False
+    return homo_score,comp_score,v_score,skip
```
shankari, line 102: I'm not going to insist on this since you are planning to refactor anyway, but you don't need to have `homo_score`, `comp_score` and `v_score` as both input and output. You can create a class that encapsulates all the scores and just pass it in directly.

Also, I really don't think you need to have append code in here - you can just have this function return valid or invalid. If invalid, append the values and continue; if valid, proceed with computation.

tour_model_eval/user_input_request_times_all_users.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "            if early_trip['data']['start_local_dt']['year']>compare_trip['data']['start_local_dt']['year']:\n",
+    "                early_trip = compare_trip\n",
+    "                index = i\n",
+    "            elif early_trip['data']['start_local_dt']['year']==compare_trip['data']['start_local_dt']['year'] and early_trip['data']['start_local_dt']['month']>compare_trip['data']['start_local_dt']['month']:\n",
+    "                early_trip = compare_trip\n",
+    "                index = i\n",
+    "            elif early_trip['data']['start_local_dt']['year']==compare_trip['data']['start_local_dt']['year'] and early_trip['data']['start_local_dt']['month']==compare_trip['data']['start_local_dt']['month'] and early_trip['data']['start_local_dt']['day']>compare_trip['data']['start_local_dt']['day']:\n",
+    "                early_trip = compare_trip\n",
+    "                index = i\n",
+    "        early_trip_index = bin[index]\n",
```
shankari, line 200: Move this into `find_first_trip`?
Also, instead of this complicated logic to compare two trips based on the `start_local_dt`, you can compare directly based on the `start_ts`. You only need `start_local_dt` to group the trips and find the number of requests per day.

tour_model_eval/user_input_request_times_all_users.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "    for bin in bl_bins:\n",
+    "        for trip_index in bin:\n",
+    "            bl_trip_ls.append(trip_index)\n",
+    "            \n",
+    "    # whole list of requested trips indices\n",
+    "    req_trips_ls=ab_trip_ls+bl_trip_ls\n",
+    "    \n",
+    "    \n",
+    "    # collect request times in a day\n",
+    "    bin_day = evaluation.bin_date(req_trips_ls,filter_trips,day=True)\n",
```
shankari, line 224: I'm confused; I don't see this method in `confirmed_trips_eval_bins_clusters.py`

shankari (changes requested):

tour_model_eval/user_input_request_times_all_users.ipynb
corinne-hcr: I don't understand. Before this step, I group the trips based on clusters, this step is to look for the first trip. But why do I need to find the number of requests per day in one cluster?

corinne-hcr: I updated this part in [644c2dd](https://github.com/e-mission/e-mission-eval-private-data/pull/19/commits/644c2dd1e64734d194270ccba0920f3bfe64d209)

shankari: I don't understand. I have a cluster with trips t1, t2, t3. I want to find the first one. Why is this not just comparing `t1.start_ts`, `t2.start_ts` and `t3.start_ts`? Why do you need complicated logic involving the year, day and month?

shankari: ah looks like you are using `start_ts` in https://github.com/e-mission/e-mission-eval-private-data/pull/19/commits/644c2dd1e64734d194270ccba0920f3bfe64d209

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -113,6 +113,66 @@ def compute_score(labels_true,labels_pred,homo_score,comp_score,v_score):
...
+                    added = True
+                    break
+
+        if not added:
+            bin_date.append([trip_index])
+
+    return bin_date
+
+
+# compare the trip orders in bin_trips with those in filter_trips above cutoff
```
shankari, line 162: Improve comment. What is "trip_orders"? why does it matter? 
Just repeating the function name in the comment doesn't add much

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -113,6 +113,66 @@ def compute_score(labels_true,labels_pred,homo_score,comp_score,v_score):
...
+                if match_day(trip,bin,filter_trips):
+                    bin.append(trip_index)
+                    added = True
+                    break
+            if month:
+                if match_month(trip,bin,filter_trips):
+                    bin.append(trip_index)
+                    added = True
+                    break
+
```
shankari, line 155: It seems like there should be a better way to implement this in pandas using `groupby`. Please file an issue for this so you can explore how to clean it up later.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
shankari: Please file an issue to clean this up later.

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -9,65 +9,327 @@
...
+    homo = metrics.homogeneity_score(labels_true, labels_pred)
+    homo_score.append(float('%.3f' % homo))
+    comp = metrics.completeness_score(labels_true, labels_pred)
+    comp_score.append(float('%.3f' % comp))
+    v = metrics.v_measure_score(labels_true, labels_pred)
+    v_score.append(float('%.3f' % v))
+    return homo_score,comp_score,v_score
+
+
+# This function is to compare a trip with a group of trips to see if they happened in a same day
```
shankari, line 116: what are the paramters?
you have a `bin` and `filter_trips`. Why do you need to pass in both?

tour_model_eval/user_input_request_times_all_users.ipynb
```diff
@@ -0,0 +1,493 @@
...
+    "    req_day_ls = []\n",
+    "    for bin in bin_day:\n",
+    "        req_day_ls.append(len(bin))\n",
+    "               \n",
```
shankari, line 153: Please add a comment here that describes what `bin_day` is expected to look like to make the rest of this easier to review

tour_model_eval/confirmed_trips_eval_bins_clusters.py
```diff
@@ -9,65 +9,327 @@
...
+                    break
+            if month:
+                if match_month(trip,bin,filter_trips):
+                    bin.append(trip_index)
+                    added = True
+                    break
+
+        if not added:
+            bin_date.append([trip_index])
+
```
shankari, line 158: This code is just so messy and non-pythonic

shankari: Please add this to the list of areas to improve the implementation.

tour_model_eval/user_input_request_times_all_users.ipynb
```diff
@@ -0,0 +1,493 @@
...
+    "    propor_single_user = []\n",
+    "    for valid_trips_bin in bin_filter_trips_day:\n",
+    "        match = False\n",
+    "        for req_trips_bin in bin_day:\n",
+    "            req_trip = filter_trips[req_trips_bin[0]]\n",
+    "            if evaluation.match_day(req_trip,valid_trips_bin,filter_trips):\n",
+    "                proportion = round(len(req_trips_bin)/len(valid_trips_bin), 2)\n",
+    "                propor_single_user.append(proportion)\n",
+    "                match = True\n",
+    "                break\n",
```
shankari, line 181: This is the problem with the current `match_day` implementation. Reading this appears that you compare one trip to a list to find the matches, and then you determine the proportion for that user? Where is the iteration across days?

This looks wrong, but it has no comments for explanation.

PR #20, opened 2021-05-12
check codes

corinne-hcr:


tour_model_eval/data_preprocessing.py
```diff
@@ -0,0 +1,54 @@
...
+
+# - trips: all trips read from database
+# - filter_trips: valid trips that have user labels and are not points
+def filter_data(user,radius):
+    trips = pipeline.read_data(uuid=user, key=esda.CONFIRMED_TRIP_KEY)
+    non_empty_trips = [t for t in trips if t["data"]["user_input"] != {}]
+    non_empty_trips_df = pd.DataFrame(t["data"]["user_input"] for t in non_empty_trips)
+    valid_trips_df = non_empty_trips_df.dropna(axis=0, how='any', thresh=None, subset=None, inplace=False)
+    valid_trips_idx_ls = valid_trips_df.index.tolist()
+    valid_trips = [non_empty_trips[i]for i in valid_trips_idx_ls]
```
shankari, line 16: why do you need `non_empty_trips` and `non_empty_trips_df`?
Isn't one of them enough?

tour_model_eval/data_preprocessing.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # similarity codes can filter out trips that are points in valid_trips
+    sim = similarity.similarity(valid_trips, radius)
+    filter_trips = sim.data
```
shankari, line 20: I gave this feedback earlier, but it is better to split `similarity.similarity` to have the filtering and binning code separately, so you can call only the filtering code here.

That is both more clear and more performant.

tour_model_eval/data_preprocessing.py
```diff
@@ -0,0 +1,54 @@
...
+
+# use KFold (n_splits=5) to split the data into 5 models (5 training sets, 5 test sets)
+def split_data(filter_trips):
+    X = []
+    for trip in filter_trips:
+        start = trip.data.start_loc["coordinates"]
+        end = trip.data.end_loc["coordinates"]
+        distance = trip.data.distance
+        duration = trip.data.duration
+        X.append([start[0], start[1], end[0], end[1], distance, duration])
```
shankari, line 32: In the classic ML paradigm, this is featurization, in which you extract features from the individual data points. This should be a separate function. Also, why are you using a list instead of a pandas dataframe.

tour_model_eval/data_preprocessing.py
```diff
@@ -0,0 +1,54 @@
...
+    train_idx = []
+    test_idx = []
+    for train_index, test_index in kf.split(X):
+        train_idx.append(train_index)
+        test_idx.append(test_index)
+    return train_idx, test_idx
+
+
+# collect a set of data(training/test set) after splitting
+# - splited_indices: pass in train set indices or test set indices
```
shankari, line 44: where is `splited_indices` defined?

tour_model_eval/data_preprocessing.py
```diff
@@ -0,0 +1,54 @@
...
+def get_subdata(sim,train_test_set):
+    collect_sub_data = []
+    for train_test_subset in train_test_set:
+        sub_data = []
```
shankari, line 48: why are you passing in the `sim` object instead of the set of valid trips? 
What if we wanted to incorporate additional filtering steps into `filter_data`?
As I indicated earlier, it is bad to rely on the `sim` code for filtering.

tour_model_eval/get_users.py
```diff
@@ -0,0 +1,31 @@
...
+def valid_user(filter_trips,trips):
+    valid = False
+    if len(filter_trips) >= 10 and len(filter_trips) / len(trips) >= 0.5:
+        valid = True
+    return valid
+
+
+# - user_ls: a list of all users
+# - valid_user_ls: a list of valid users
+def get_user_ls(all_users,radius):
```
shankari, line 15: how is `user_ls` different from `all_users`?

tour_model_eval/get_users.py
```diff
@@ -0,0 +1,31 @@
...
+    user_ls = []
+    valid_user_ls = []
+    for i in range(len(all_users)):
+        curr_user = 'user' + str(i + 1)
+        user = all_users[i]
+        filter_trips,sim,trips = preprocess.filter_data(user,radius)
```
shankari, line 21: where do you use `sim` here? Again, a reason to clean up the interface for `filter_data`

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,98 @@
...
+from scipy.cluster.hierarchy import linkage, dendrogram,fcluster
+
+
+
+# to map the user labels
+# - user_input_df: pass in original user input dataframe, return changed user input dataframe
+# - sp2en: change Spanish to English
+# - cvt_pur_mo: convert purposes and replaced mode
+def map_labels(user_input_df,sp2en,cvt_pur_mo):
+    # Spanish words to English
```
shankari, line 11: My gut  check is to split this into two functions - one for `sp2en` and one for `cvt_pur_mo` instead of smushing them into one giant function. What is the advantage of combining them?

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,98 @@
+import logging
+from scipy.cluster.hierarchy import linkage, dendrogram,fcluster
```
shankari, line 2: in terms of code style, I prefer 
`import x.y.z as xyz` instead of `from x.y.z import foo`
that makes it clear where you are importing functions from

Please change.

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+# use hierarchical clustering to get labels of the second round
+def get_second_labels(x,method,low,dist_pct):
+    z = linkage(x, method=method, metric='euclidean')
+    last_d = z[-1][2]
+    clusters = []
+    if last_d < low:
+        for i in range(len(x)):
+            clusters.append(0)
```
shankari, line 46: please add a comment around what this does with an example.

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,98 @@
...
+        clusters = fcluster(z, max_d, criterion='distance')
+    return clusters
+
+
+# this function includes hierarchical clustering and changing labels to get appropriate labels for
+# the second round of clustering
+# appropriate labels are label from the first round concatenate label from the second round
+# (e.g. label from first round is 1, label from second round is 2, the new label will be 12)
+def get_new_labels(x,low,dist_pct,second_round_idx_labels,new_labels,method=None):
+    idx_label = second_round_idx_labels.copy()
```
shankari, line 58: what is `second_round_idx_labels` and why do you need to `copy()` it?

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,98 @@
...
+        for i in range(len(bin)):
+            bin[i] = track[bin[i]][0]
+    return bin_sim_trips
+
+
+# replace the first round labels with new labels
+def change_track_labels(track,new_labels):
+    for i in range(len(new_labels)):
+        track[i][1] = new_labels[i]
+    return track
```
shankari, line 95: add some documentation on what is a `track`

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,98 @@
...
+    bin_sim_trips = []
+    for trip_index,label in enumerate(new_labels):
+        added = False
+        for bin in bin_sim_trips:
+            if label == new_labels[bin[0]]:
+                bin.append(trip_index)
+                added = True
+                break
+        if not added:
+            bin_sim_trips.append([trip_index])
```
shankari, line 83: you should be able to replace this with a much more readable and performant function. Is a list really the best way to deal with this? Would a dict `{label : list}` be a better option? If a list is the only option, can you use `find` or `filter` functions instead of a manual iteration?

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,98 @@
...
+    # using track to replace the current indices with original indicies
+    for bin in bin_sim_trips:
+        for i in range(len(bin)):
+            bin[i] = track[bin[i]][0]
+    return bin_sim_trips
```
shankari, line 88: again, not  clear what `track` represents or what you are trying to achieve here

shankari (changes requested):
Several changes around modularity and documentation.

tour_model_eval/data_preprocessing.py
shankari: Move the current filtering implementation in similarity into a separate function called `filter_distance`.
- Call `filter_distance` from your code
- Change the `similarity.similarity` first step to call `filter_distance`

shankari: Move the current filtering implementation in similarity into a separate function called `filter_distance`.
- Call `filter_distance` from your code
- Change the `similarity.similarity` first step to call `filter_distance`

tour_model_eval/data_preprocessing.py
```diff
@@ -0,0 +1,54 @@
+import emission.storage.decorations.analysis_timeseries_queries as esda
+import emission.analysis.modelling.tour_model.cluster_pipeline as pipeline
+import emission.analysis.modelling.tour_model.similarity as similarity
+import pandas as pd
+from sklearn.model_selection import KFold
+
+
+# - trips: all trips read from database
+# - filter_trips: valid trips that have user labels and are not points
+def filter_data(user,radius):
```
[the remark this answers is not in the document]
corinne-hcr: changed

tour_model_eval/data_preprocessing.py
corinne-hcr: deleted

tour_model_eval/data_preprocessing.py
corinne-hcr: changed

tour_model_eval/get_users.py
corinne-hcr: `user_ls` is like user1, user1
`all_users` is user id

tour_model_eval/get_users.py
corinne-hcr: I need to run `sim` functions after getting `filter_trip`. Maybe it is better to show in a more complete pipeline

tour_model_eval/label_processing.py
corinne-hcr: changed

tour_model_eval/label_processing.py
corinne-hcr: done

tour_model_eval/label_processing.py
corinne-hcr: done

tour_model_eval/label_processing.py
corinne-hcr: It is a track to store the indices and labels for the second round.

tour_model_eval/label_processing.py
corinne-hcr: done

tour_model_eval/label_processing.py
corinne-hcr: done

tour_model_eval/label_processing.py
```diff
@@ -1,65 +1,77 @@
...
+    user_input_df = map_labels_sp2en(user_input_df)
+    # convert purpose
+    user_input_df = user_input_df.replace(map_pur_dict)
+    # convert mode
+    for a in range(len(user_input_df)):
+        if user_input_df.iloc[a]["replaced_mode"] == "same_mode":
+            # to see which row will be converted
+            logging.debug("The following rows will be changed: %s", user_input_df.iloc[a])
+            user_input_df.iloc[a]["replaced_mode"] = user_input_df.iloc[a]['mode_confirm']
     return user_input_df
```
shankari: I would pull this out into a separate function.
It looks like you have three different cleanup steps.
Instead of putting them all into one or two functions, have `map_labels_sp2en` `map_labels_pur` and `map_labels_mode` separately. Then call them all in sequence in the pipeline. That makes it easier to experiment with (for example) only changing the mode or the purpose.

tour_model_eval/label_processing.py
```diff
@@ -1,65 +1,77 @@
...
-        clusters = fcluster(z, max_d, criterion='distance')
+        clusters = sch.fcluster(z, max_d, criterion='distance')
     return clusters
 
 
 # this function includes hierarchical clustering and changing labels to get appropriate labels for
 # the second round of clustering
 # appropriate labels are label from the first round concatenate label from the second round
 # (e.g. label from first round is 1, label from second round is 2, the new label will be 12)
+# - second_round_idx_labels: a track to store the indices and labels for the second round
```
shankari, line 69: what does "a track" mean?

tour_model_eval/label_processing.py
```diff
@@ -1,65 +1,77 @@
...
 def get_new_labels(x,low,dist_pct,second_round_idx_labels,new_labels,method=None):
-    idx_label = second_round_idx_labels.copy()
     second_labels = get_second_labels(x,method,low,dist_pct)
     for i in range(len(second_labels)):
-        index = idx_label[i][0]
-        new_label = idx_label[i][1]
+        index = second_round_idx_labels[i][0]
+        new_label = second_round_idx_labels[i][1]
```
shankari, line 74: clarify what is the `index` and the `second_label`. If the `index` is the first label (which is what the code seems to show), then calling it `first_index` would be good 😄

tour_model_eval/label_processing.py
```diff
@@ -89,6 +101,10 @@ def group_similar_trips(new_labels,track):
 
 
 # replace the first round labels with new labels
+# - track: a list to store the indices and labels from the first round of clustering
+# for item in track, item[0] is the original index of the trip in filter_trips
+# item[1] is the label after the first round of clustering
+# we change the labels from the first round with new labels from the second round here
```
shankari, line 107: If this gets more complicated, it may be worthwhile to create a new class called `track` so this is encapsulated better.

tour_model_eval/data_preprocessing.py
shankari: As an example of how to merge trip information and user label information into the same table, please look at
https://github.com/e-mission/em-public-dashboard/blob/main/viz_scripts/scaffolding.py#L61

tour_model_eval/get_users.py
shankari: Add a comment here for others to read as well

tour_model_eval/label_processing.py
shankari: what do you mean by "the list of clusters"? Do you mean "list of labels"?
And would the labels always be zero, or would they depend on the first round labels?

tour_model_eval/label_processing.py
shankari: please add a note here referring to what "track" means

shankari (changes requested):
The next step is to clean up/refactor the similarity code.
There are also some minor/comment-related changes to address.
Let's leave the list -> dataframe change until later.

tour_model_eval/label_processing.py
corinne-hcr: done

tour_model_eval/label_processing.py
corinne-hcr: changed

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,128 @@
...
+            user_input_df.iloc[a]["replaced_mode"] = user_input_df.iloc[a]['mode_confirm']
+    return user_input_df
+
+
+# this function will change Spanish to English, convert purposes, and convert modes
+def map_labels(user_input_df):
+    user_input_df = map_labels_purpose(user_input_df)
+    user_input_df = map_labels_mode(user_input_df)
+    return user_input_df
+
```
shankari, line 49:
```suggestion
def map_labels(user_input_df):
    # Note that the spanish -> english conversion MUST currently happen before the other
    # mode and purpose mappings
    user_input_df = map_labels_sp2en(user_input_df)
    user_input_df = map_labels_purpose(user_input_df)
    user_input_df = map_labels_mode(user_input_df)
    return user_input_df
```

tour_model_eval/get_users.py
```diff
@@ -0,0 +1,32 @@
...
+# - user_ls: a list of all users, like [user1, user2, user3...]
+# - valid_user_ls: a list of valid users
+# - all_users: the collection of all user id
+def get_user_ls(all_users,radius):
+    user_ls = []
+    valid_user_ls = []
+    for i in range(len(all_users)):
+        curr_user = 'user' + str(i + 1)
+        user = all_users[i]
+        filter_trips,sim,trips = preprocess.filter_data(user,radius)
```
shankari, line 22: Don't you need to remove this at a minimum? `filter_data` no longer returns 3 values.

```suggestion
        filter_trips,trips = preprocess.filter_data(user,radius)
```

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,128 @@
...
+    return clusters
+
+
+# this function includes hierarchical clustering and changing labels from the first round to get appropriate labels for
+# the second round of clustering
+# appropriate labels are label from the first round concatenate label from the second round
+# (e.g. label from first round is 1, label from second round is 2, the new label will be 12)
+# - second_round_idx_labels: a list to store the indices and labels from the first round.
+# - second_labels: labels from the second round of clustering
+# We will use it as a "track" in the following.
```
shankari, line 83:
```suggestion
# We will use it as a "track" in the following. For more information on the "track" data structure, please see the `change_track_labels` function

```

tour_model_eval/get_users.py
```diff
@@ -0,0 +1,32 @@
...
+def valid_user(filter_trips,trips):
+    valid = False
+    if len(filter_trips) >= 10 and len(filter_trips) / len(trips) >= 0.5:
+        valid = True
+    return valid
+
+
+# - user_ls: a list of all users, like [user1, user2, user3...]
+# - valid_user_ls: a list of valid users
+# - all_users: the collection of all user id
```
shankari, line 15:
```suggestion
# - user_ls: a list of strings representing short user names, such as [user1, user2, user3...]
# - valid_user_ls: a subset of `user_ls` for valid users, so also string representation of user names
# - all_users: a collection of all user ids, in terms of user id objects
```

shankari (changes requested):
My comments are as suggestions now since we have now gone through two rounds of feedback. There are still pending changes, IMHO.

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,128 @@
...
+
+# to map purposes and replaced mode in user inputs
+# - cvt_pur_mo: convert purposes and replaced mode
+def map_labels_purpose(user_input_df):
+    # Convert purpose
+    map_pur_dict = {'course': 'school', 'work_- lunch break': 'lunch_break', 'on_the way home': 'home',
+                    'insurance_payment': 'insurance'}
+
+    # change language first
+    user_input_df = map_labels_sp2en(user_input_df)
```
shankari, line 28: Since we are calling this directly from `map_labels`
```suggestion
```

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,126 @@
...
+    user_input_df = map_labels_mode(user_input_df)
+    return user_input_df
+
+# use hierarchical clustering to get labels of the second round
+# - sch.linkage: perform hierarchical(agglomerative) clustering
+# In this function, we set a low bound and a higher bound(cutoff) of distance in the dendrogram
+# - last_d: the distance of the last cluster in the dendrogram
+# - low: the lower bound of distance
+# e.g., if low = 300, last_d = 250, we will assign 0s as labels for the points,
+# and the list of labels will be like [0,0,0,0,0].
```
shankari, line 57:
```suggestion
# e.g., if low = 300, last_d = 250, we will assign 0s as labels for the points, irrespective of the first round labels.
# and the list of second round labels will be like [0,0,0,0,0].
```

shankari (changes requested):

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # find the unique set of bins and store their indices into `bin_sim_trips`
+    label_set = set(new_labels)
+    for sel_label in label_set:
+        # convert the set of unique labels into their indices
+        # concretely, if the input labels are ['a','a','a','b','b','b']
+        # the unique labels are ['a', 'b']
+        # the bin array is [0,0,0,1,1,1]
+        bin = [index for (index, label) in enumerate(new_labels) if label == sel_label]
+        bin_sim_trips.append(bin)
```
shankari, line 109:
```suggestion
    label_set = set(new_labels)
    # convert the set of unique labels into their indices
    # concretely, if the input labels are ['a','a','a','b','b','b']
    # the unique labels are ['a', 'b']
    for sel_label in label_set:
        # for the first iteration, bin = [0,1,2]
        # for the second iteration, bin = [3,4,5]
        bin = [index for (index, label) in enumerate(new_labels) if label == sel_label]
        bin_sim_trips_idx.append(bin)
    # At the end, bin_sim_trips_idx = [[0,1,2],[3,4,5]]
```

tour_model_eval/label_processing.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # using track to replace the current indices with original indices
+    for bin in bin_sim_trips:
+        for i in range(len(bin)):
+            bin[i] = track[bin[i]][0]
+    return bin_sim_trips
```
shankari, line 115:
```suggestion
    # using track to replace the current indices with original indices
    for bin in bin_sim_trips_idx:
        # in the first iteration, bin = [0,1,2]
        # in the first iteration of that, we map the trip index of the
        # common trip (e.g. 0) to the original index for that trip from the track (e.g. 42)
        for i in range(len(bin)):
            bin[i] = track[bin[i]][0]
    # At this point, the bin_sim_trips_idx will have original indices for the trips
    return bin_sim_trips_idx
```

tour_model_eval/data_preprocessing.py
```diff
@@ -0,0 +1,58 @@
...
+
+# use KFold (n_splits=5) to split the data into 5 models (5 training sets, 5 test sets)
+def split_data(filter_trips):
+    X = []
+    for trip in filter_trips:
+        start = trip.data.start_loc["coordinates"]
+        end = trip.data.end_loc["coordinates"]
+        distance = trip.data.distance
+        duration = trip.data.duration
+        X.append([start[0], start[1], end[0], end[1], distance, duration])
```
shankari, line 37:
```suggestion
def extract_features(filter_trips):
    X = []
    for trip in filter_trips:
        start = trip.data.start_loc["coordinates"]
        end = trip.data.end_loc["coordinates"]
        distance = trip.data.distance
        duration = trip.data.duration
        X.append([start[0], start[1], end[0], end[1], distance, duration])
```

tour_model_eval/data_preprocessing.py
```diff
@@ -0,0 +1,58 @@
...
+
+    kf = KFold(n_splits=5, shuffle=True, random_state=3)
+    train_idx = []
+    test_idx = []
+    for train_index, test_index in kf.split(X):
+        train_idx.append(train_index)
+        test_idx.append(test_index)
+    return train_idx, test_idx
```
shankari, line 45:
```suggestion
def split_data(X):
    kf = KFold(n_splits=5, shuffle=True, random_state=3)
    train_idx = []
    test_idx = []
    for train_index, test_index in kf.split(X):
        train_idx.append(train_index)
        test_idx.append(test_index)
    return train_idx, test_idx
```

PR #21, opened 2021-05-27
check codes

corinne-hcr:


tour_model_eval/get_request_percentage.py
```diff
@@ -0,0 +1,105 @@
...
+    # collect common trip indices above cutoff
+    no_req_trip_ls = []
+    for bin in new_bins:
+        early_trip_index, index = find_first_trip(filter_trips, bin)
+        ab_trip_ls.append(early_trip_index)
+
+        for k in range(len(bin)):
+            if k != index:
+                no_req_trip_idx = bin[k]
+                no_req_trip_ls.append(no_req_trip_idx)
```
shankari, line 74: again, what is the difference between `early_trip_index` and `index`?
what are you doing in this for loop and can you either replace it with something simpler, or add some additional explanations here? What does `no_req_trip_idx = bin[k]` even do?

tour_model_eval/get_request_percentage.py
```diff
@@ -0,0 +1,105 @@
...
+    return req_trips_ls
+
+
+# get request percentage based on the number of requested trips and the total number of trips
+def get_req_pct(new_labels,track,filter_trips):
+    # - new_bins: bins with original indices of similar trips
+    new_bins = label_pro.group_similar_trips(new_labels,track)
+    req_trips = get_requested_trips(new_bins,filter_trips,sim)
+    pct = len(req_trips)/len(filter_trips)
+    return pct
```
shankari, line 105: I'm not quite sure what you are doing here. If you just want the number of requested trips and total number of trips, why do you have to change the labels?

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+
+# compare the trip orders in bin_trips with those in filter_trips above cutoff
+def compare_trip_orders(bins,bin_trips,filter_trips):
+    bin_trips_ts = pd.DataFrame(data=[trip["data"]["start_ts"] for trip in bin_trips])
+    bin_ls = []
+    for bin in bins:
+        for index in bin:
+            bin_ls.append(index)
```
shankari, line 14: why do you need to iterate like this? Why not just use the bins directly? Is it to allow us to `flatten`?

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # collect labels_true based on user_input
+    labels_true = []
+    for trip in bin_trips_user_input_ls:
+        if trip in no_dup_list:
+            labels_true.append(no_dup_list.index(trip))
+
+    labels_pred = labels_pred
+    homo_score = metrics.homogeneity_score(labels_true, labels_pred)
+    return homo_score
```
shankari, line 42: I'm a little confused here about how you compute `labels_true`, you drop duplicates from the input list before creating `labels_true`. Why? If there are multiple labels in the bin, why would we not keep all of them? Why don't you need to drop duplicates on `labels_pred`?

Can you walk through how this function would work for some sample inputs?

tour_model_eval/get_plot.py
```diff
@@ -0,0 +1,54 @@
...
+    x=percentage
+    y=homo_score
+    v=valid_users
+    cmp = cm.get_cmap('Dark2', len(valid_users))
+
+    sc = []
+    for i in range(len(valid_users)):
+        for n in range(len(x[i])):
+            point = plt.scatter(x[i][n], y[i][n], color=cmp.colors[i], s=70, alpha=0.7)
+        sc.append(point)
```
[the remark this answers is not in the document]
corinne-hcr: I don't understand what you mean by using a dataframe or a numpy array here. Since we will have several points for a user, in order to make one legend for one user, I need to collect one of the several points of that user, and `i` here is to control the color. I don't know how to use  a numpy array to do so.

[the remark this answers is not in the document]
shankari: did you try to search for "pandas dataframe scatter plot"?
One of the first hits is https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.scatter.html

tour_model_eval/get_request_percentage.py
```diff
@@ -0,0 +1,105 @@
...
+def requested_trips_bl_cutoff(sim):
+    # bins below cutoff
+    bl_bins = sim.below_cutoff
+
+    # collect requested trips indices below cutoff
+    bl_trip_ls = []
+    for bin in bl_bins:
+        for trip_index in bin:
+            bl_trip_ls.append(trip_index)
+    return bl_trip_ls
```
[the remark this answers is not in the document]
corinne-hcr: The point for this is to collect the flatten indices of trips below cutoff (uncommon trips), and then I can concatenate the requested trips from common trips.
I don't understand what you mean by using `flatten`. I see other people also write a similar function to get a flatten list.

tour_model_eval/get_request_percentage.py
corinne-hcr: I didn't changed the labels here. I just put trips with same labels together.

tour_model_eval/get_request_percentage.py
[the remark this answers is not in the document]
shankari: if your code is flattening a list, it helps to say that so people don't need to think through the loop and figure out what is going on - e.g. 

Note also that there are other one-line functions for flattening - e.g. https://stackoverflow.com/questions/952914/how-to-make-a-flat-list-out-of-a-list-of-lists

```suggestion
    # effectively, bl_trip_ls = flatten(bl_bins)
    # >>> bl_bins = [[1,2],[3,4],[5,6]]
    # >>> bl_trip_ls = [item for sublist in bl_bins for item in sublist]
    # >>> bl_trip_ls
    # [1, 2, 3, 4, 5, 6]
    # XXX: add here the reason why you want to flatten the list
    bl_trip_ls = [item for sublist in bl_bins for item in sublist]
```

tour_model_eval/get_request_percentage.py
shankari: The variable is called `new_labels` so it sounds like there are some **new** labels. if not, can you clarify what the input is? Again, an example of what a "track" might look like, what "new_labels" looks like, and what this code does would be helpful.

tour_model_eval/get_request_percentage.py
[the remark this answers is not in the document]
corinne-hcr: Oh,ok. I did find a function online, but haven't test it yet on  `bl_trip_ls`
for example, if I want to flatten `bins`, I use bin_ls = list(np.concatenate(bins).flat)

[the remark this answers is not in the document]
shankari: But honestly, either of them would be fine, just to make clear what you are doing there.

tour_model_eval/get_request_percentage.py
```diff
@@ -0,0 +1,105 @@
...
+def find_first_trip(filter_trips,bin):
+    early_trip = filter_trips[bin[0]]
+    index = 0
+    for i in range(1,len(bin)):
+        compare_trip = filter_trips[bin[i]]
+        if early_trip['data']["start_ts"] > compare_trip['data']["start_ts"]:
+            early_trip = compare_trip
+            index = i
+    early_trip_index = bin[index]
+    return early_trip_index, index
```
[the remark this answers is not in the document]
corinne-hcr: resolved

tour_model_eval/get_request_percentage.py
corinne-hcr: resolved

tour_model_eval/get_request_percentage.py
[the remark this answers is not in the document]
corinne-hcr: done, I copy your suggestion in my code and commit them with other changes.

tour_model_eval/get_scores.py
corinne-hcr: resolved

tour_model_eval/get_request_percentage.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def find_first_trip(filter_trips,bin):
+    early_trip = filter_trips[bin[0]]
+    index = 0
+    for i in range(1,len(bin)):
+        compare_trip = filter_trips[bin[i]]
+        if early_trip['data']["start_ts"] > compare_trip['data']["start_ts"]:
+            early_trip = compare_trip
+            index = i
```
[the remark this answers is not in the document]
corinne-hcr: resolved

tour_model_eval/get_request_percentage.py
```diff
@@ -0,0 +1,117 @@
...
+        early_trip_index, early_idx_in_bin = find_first_trip(filter_trips, bin)
+        ab_trip_ls.append(early_trip_index)
+
+        # The following loop collects the original indices of the rest of the trips in the bin. Since they are not the
+        # earliest one, we don't need to request for user labels
+        for k in range(len(bin)):
+            if k != early_idx_in_bin:
+                # - no_req_trip_idx: the original index of the trip that is not requested for user labels
+                no_req_trip_idx = bin[k]
+                no_req_trip_ls.append(no_req_trip_idx)
```
shankari, line 75: I still don't see why you need to do this. You already have the index of the trip, so why can't you just do `if bin[k] != early_trip_idx`? Or better still 

```suggestion
        # The following loop collects the original indices of the rest of the trips in the bin. Since they are not the
        # earliest one, we don't need to request for user labels
        # >>> x = [100,200,300]
        # >>> x.remove(100); x
        # [200, 300]
        no_req_trip_ls = copy.copy(bin)
        no_req_trip_ls = bin.remove(early_trip_index)
```

shankari (changes requested):
This is an incomplete review. I will continue today evening.

tour_model_eval/get_request_percentage.py
```diff
@@ -0,0 +1,117 @@
...
+            if month:
+                if match_month(trip,bin,filter_trips):
+                    bin.append(trip_index)
+                    added = True
+                    break
+
+        if not added:
+            bin_date.append([trip_index])
+
+    return bin_date
```
shankari, line 47: All this code needs to be replaced by `dataframe.groupby` in the future

shankari: @corinne-hcr please update the cleanup issue with a full list of changes that need to happen after this weekend.

tour_model_eval/get_request_percentage.py
```diff
@@ -0,0 +1,117 @@
...
+    # bins below cutoff
+    bl_bins = sim.below_cutoff
+
+    # collect requested trips indices below cutoff
+    # effectively, bl_trip_ls = flatten(bl_bins)
+    # >>> bl_bins = [[1,2],[3,4],[5,6]]
+    # >>> bl_trip_ls = [item for sublist in bl_bins for item in sublist]
+    # >>> bl_trip_ls
+    # [1, 2, 3, 4, 5, 6]
+    # the reason for flattening: add here the reason why you want to flatten the list
```
shankari, line 90: As the comment says, please add here the reason for why you want to flatten the list.

tour_model_eval/get_request_percentage.py
```diff
@@ -0,0 +1,117 @@
...
+    # - new_label: For the first round, new_label is the copy of the first round labels, e.g. [1,1,1,2,2,2].
+    # For the second round, new_label is that the first round label concatenate the second round label.
+    # e.g.the label from the second round is [1,2,1,2,3,3], new_label will turn to [11,12,11,22,23,23]
+    # - track: at this point, each item in the track contains the original index of a trip,
+    # and the latest label of it. e.g. [ori_idx, latest_label]
+    # concretely, please look at "group_similar_trips" function in label_processing.py
+    # If new_label is [11,12,11,22,23,23] and the original indices of the trips is [1,2,3,4,5,6],
+    # new_bins will be [[1,3],[2],[4],[5,6]]
+    new_bins = label_pro.group_similar_trips(new_labels,track)
+    req_trips = get_requested_trips(new_bins,filter_trips,sim)
```
shankari, line 115: but if you already have the `new_bins`, why do you need to pass in `sim` and `filter_trips`? That is the part that is confusing me. `get_requested_trips` returns the trips above and below the cutoff, I understand that. But `sim` is also related to the (old) bins and `filter_trips` is just all trips. Why do you need those?

In particular, since you are working with indices everywhere, I don't see any reason why you need to pass in filter_trips, which (I presume) is a set of filter objects.

tour_model_eval/get_plot.py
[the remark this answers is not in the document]
shankari: Please document that this needs to be cleaned up later (https://github.com/e-mission/e-mission-docs/issues/631)

shankari (changes requested):
That's all my comments. I will review again today evening at around 5pm. Please address the comments by then.

tour_model_eval/get_request_percentage.py
corinne-hcr: Why do you remove `early_trip_index` from `bin`? I think you use `no_req_trip_ls` as a copy of `bin`, why not remove from `no_req_trip_ls` ?

shankari: Good point, I meant `no_req_trip_ls.remove`

tour_model_eval/get_request_percentage.py
corinne-hcr: I added some comment in the code. Please see if they are clear to you.

tour_model_eval/get_scores.py
corinne-hcr: done

tour_model_eval/get_request_percentage.py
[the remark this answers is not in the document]
shankari: Where is this resolved? This code looks unchanged to me.

tour_model_eval/get_request_percentage.py
```diff
@@ -0,0 +1,131 @@
...
+
+        # The following loop collects the original indices of the rest of the trips in the bin. Since they are not the
+        # earliest one, we don't need to request for user labels
+        # >>> x = [100,200,300]
+        # >>> x.remove(100); x
+        # [200, 300]
+        no_req_trip_subls = copy.copy(bin)
+        no_req_trip_subls.remove(early_trip_index)
+        no_req_trip_ls.append(no_req_trip_subls)
+    no_req_trip_ls = list(itertools.chain(*no_req_trip_ls))
```
shankari, line 82: you can do the following to keep the code similar to `ab_trip_ls`, right?

```suggestion
        # >>> x = [1,2,3]
        # >>> x.extend([4,5,6]); x
        # [1, 2, 3, 4, 5, 6]
        no_req_trip_ls.extend(no_req_trip_subls)
```

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,44 @@
...
+def compare_trip_orders(bins,bin_trips,filter_trips):
+    bin_trips_ts = pd.DataFrame(data=[trip["data"]["start_ts"] for trip in bin_trips])
+    bin_ls = list(itertools.chain(*bins))
+    bins_ts = pd.DataFrame(data=[filter_trips[i]["data"]["start_ts"] for i in bin_ls])
+    # compare two data frames, the program will continue to score calculation if two data frames are the same
+    assert_frame_equal(bins_ts, bin_trips_ts)
+
+
+# This function is to get homogeneity score after the first/second round of clustering
+# It is based on bin_trips, which are common trips. bin_trips are collected according to the indices in bins
```
shankari, line 18: indices in the bins, or indices of the trips?

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,44 @@
+import pandas as pd
+from pandas.testing import assert_frame_equal
+import label_processing as label_pro
+from sklearn import metrics
```
shankari, line 4:
```suggestion
import sklearn.metrics as skm
```

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,44 @@
+import pandas as pd
+from pandas.testing import assert_frame_equal
```
shankari, line 2:
```suggestion
import pandas.testing as pdt
```

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,44 @@
...
+import itertools
+
+
+# compare the trip orders in bin_trips with those in filter_trips above cutoff
+def compare_trip_orders(bins,bin_trips,filter_trips):
+    bin_trips_ts = pd.DataFrame(data=[trip["data"]["start_ts"] for trip in bin_trips])
+    bin_ls = list(itertools.chain(*bins))
+    bins_ts = pd.DataFrame(data=[filter_trips[i]["data"]["start_ts"] for i in bin_ls])
+    # compare two data frames, the program will continue to score calculation if two data frames are the same
+    assert_frame_equal(bins_ts, bin_trips_ts)
```
shankari, line 14:
```suggestion
    pdt.assert_frame_equal(bins_ts, bin_trips_ts)
```

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,44 @@
...
+    # if user labels are [home, work, home, home, work, exercise], the unique label list is [0,1,2],
+    # labels_true will be [0,1,0,0,1,2]
+    # labels_pred is the flattened list of labels of all common trips, e.g.[1,1,11,12,13,22,23]
+    labels_true = []
+    for trip in bin_trips_user_input_ls:
+        if trip in no_dup_list:
+            labels_true.append(no_dup_list.index(trip))
+
+    labels_pred = labels_pred
+    homo_score = metrics.homogeneity_score(labels_true, labels_pred)
```
shankari, line 43:
```suggestion
    homo_score = skm.homogeneity_score(labels_true, labels_pred)
```

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,44 @@
...
+    no_dup_df = bin_trips_user_input_df.drop_duplicates()
+    # turn non-duplicate user_input into list
+    no_dup_list = no_dup_df.values.tolist()
+
+    # collect labels_true based on user_input
+    # To compute labels_true, we need to find out non-duplicate user labels, and use the index of the unique user label
+    # to label the whole trips
+    # if user labels are [home, work, home, home, work, exercise], the unique label list is [0,1,2],
+    # labels_true will be [0,1,0,0,1,2]
+    # labels_pred is the flattened list of labels of all common trips, e.g.[1,1,11,12,13,22,23]
```
shankari, line 36: I still don't understand this. Don't the labels in `labels_true` and `labels_pred` have to be the same for the homogeneity score to work?

I'm actually really unsure about how this score works. This seems to indicate that the labels don't matter as long as there is a unique mapping between them.

```
>>> import sklearn.metrics as skm
>>> skm.homogeneity_score(["A", "B", "C", "B", "C"], [0,1,2,1,2])
1.0
```

But wait, no matter what I do, I get 1.0? I would like to see some examples of how the score calculation works, maybe in an auxiliary notebook that can eventually get converted to unit tests?

```
>>> skm.homogeneity_score(["A", "B", "C", "B", "C"], [0,1,2,1,3])
1.0
>>> skm.homogeneity_score(["A", "B", "C", "B", "C"], [0,1,2,1,4])
1.0
```

shankari: It also seems like it would be much simpler if you had the same labels for `labels_pred` and `labels_true`

shankari: and if the actual labels don't matter anyway, why do you need to convert 'home','work' etc to Yet Another set of indices?

shankari: Same behavior if you change the first set of labels as well

```
>>> skm.homogeneity_score(["A", "B", "D", "B", "C"], [0,1,10,1,4])
1.0
>>> skm.homogeneity_score(["A", "B", "C", "B", "C"], [0,1,10,1,4])
1.0
```

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,44 @@
...
+    labels_true = []
+    for trip in bin_trips_user_input_ls:
+        if trip in no_dup_list:
+            labels_true.append(no_dup_list.index(trip))
```
shankari, line 40: this is not really "trip", it is "userinput", right? I think this is what you mean.

```suggestion
    for userinput_dict in bin_trips_user_input_ls:
        if userinput_dict in no_dup_list:
            labels_true.append(no_dup_list.index(userinput_dict))
```

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,44 @@
...
+    bin_trips_user_input_ls = bin_trips_user_input_df.values.tolist()
+    # drop duplicate user_input
+    no_dup_df = bin_trips_user_input_df.drop_duplicates()
+    # turn non-duplicate user_input into list
+    no_dup_list = no_dup_df.values.tolist()
+
+    # collect labels_true based on user_input
+    # To compute labels_true, we need to find out non-duplicate user labels, and use the index of the unique user label
+    # to label the whole trips
+    # if user labels are [home, work, home, home, work, exercise], the unique label list is [0,1,2],
```
shankari, line 34: the user labels are not individual strings, but tuples, correct? Please clarify to avoid confusion.

tour_model_eval/get_request_percentage.py
shankari: Maybe clarify that `new_bins`represents *above_cutoff* bins only.

shankari (changes requested):
Almost there.

tour_model_eval/get_scores.py
corinne-hcr: `labels_true` and `labels_pred` have to be numeric. It doesn't work when you pass in non-numeric values. I think that is all ML algorithms requires, right?
[Here is the documentation.](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.homogeneity_score.html)
I don't think we should explain everything in the py file that's well documented. 
It says - This metric is independent of the absolute values of the labels: a permutation of the class or cluster label values won’t change the score value in any way.
I have tried the labels and score for many times with examples when I learned how they used the labels to compute the score. There is no way to make `labels_pred` and `labels_true` with same labels. We have [home, work, exercise] as true user labels. But we may have [0,1,2] or [3,4,5] or other numeric labels after clustering. We don't need to make them exactly the same. `homogeneity_score` only sees the position of the labels.

shankari: > It doesn't work when you pass in non-numeric values.

I don't find any evidence for this. It seems to work just fine with strings, as I showed above. Can you show that it doesn't work with strings?

shankari: > There is no way to make `labels_pred` and `labels_true` with same labels.

I don't believe that it is impossible.
And if it were, we should abandon the project right now because eventually, we do need to make `labels_pred` map to the actual user labels.
It is not every useful to tell users that their trip was in cluster 5.
We have to tell them that the trip was for `home` or `work` or ....

shankari: >  It says - This metric is independent of the absolute values of the labels: a permutation of the class or cluster label values won’t change the score value in any way.

I did consider the use case of a permutation - that is the first test that I did. What I don't understand is this

```
>>> skm.homogeneity_score(["A", "B", "C", "B", "C"], [0,1,2,1,4])
1.0
```

Label `C` maps to both `2` and `4`, but the score is still 1.0

corinne-hcr: > Label C maps to both 2 and 4, but the score is still 1.0
If that is indeed the behavior of the homogeneity score, I am not sure it is a good metric for our use case.
Or at least, it needs stronger justification that it captures what we are looking for.

For this, the documentation has clearly explained.

> Non-perfect labelings that further split classes into more clusters can be perfectly homogeneous:
```
>>> print("%.6f" % homogeneity_score([0, 0, 1, 1], [0, 0, 1, 2]))
1.000000
>>> print("%.6f" % homogeneity_score([0, 0, 1, 1], [0, 1, 2, 3]))
1.000000
```
In your case, "C" is in index 2 and 4. In your `labels_pred`, index 2 and 4 can have labels 2 and 4 or 3 and 3. If they have same labels, that means they are in the same cluster, if not, it is still fine ( it means you have different clusters for same trips, e.g. 2 `home` trips with same modes in 2 clusters instead of in 1 cluster according to the Euclidean Distance). But if index 3, which is the same position as the second "B" in your `label_true`, also has label 2, the score will not be 1.0.
```
>>> skm.homogeneity_score(["A", "B", "C", "B", "C"], [0,1,2,2,4])
0.737175493807016
```

shankari: @corinne-hcr I agree that the homogeneity score documentation has the explanation, but **you need to map it to this problem and justify the usage of this metric**. The explanation above about the home trips into two clusters is what I was looking for. I would also add the explanation of how we do capture that difference in our case (home trips split into two clusters will have the same homogeneity score but a lower request percentage)

Also, all your explanations above have focused on a single string - e.g. "home", but the user input is a tuple. You need to indicate how you are dealing with that.

corinne-hcr: For mapping the labels, in one cluster, we do have trips with user labels. There are only two situations - trips are grouped in one cluster or the single trip in one cluster. Ideally, we should have labels [11,11,11] for trips in one cluster after the second round. In such a cluster, we should already have 1 or 2 trips that have user labels, the third one is predicted. Then we label it with the same user labels as the previous two. We don't have 1.0 homogeneity score for most of the cluster. It can also be [11,11,12]. Then for the third one, we will still label it with the same user labels as the previous two (since they are in one cluster).

shankari: I don't see any of this explanation in the comments. Note that if people look at the code, they don't necessarily know which PR it came from and what the history in there is. You can add a link to this discussion if you like, but there *must* be at least a link to a high level description in a code comment.

Also, I don't see a follow up to my previous comments about labeling.
https://github.com/e-mission/e-mission-eval-private-data/pull/21#discussion_r642009487
https://github.com/e-mission/e-mission-eval-private-data/pull/21#discussion_r642009788

shankari:
@corinne-hcr I don't see any comment updates all day today?

tour_model_eval/get_scores.py
```diff
@@ -15,8 +15,20 @@ def compare_trip_orders(bins,bin_trips,filter_trips):
...
+# in bins above cutoff
 # More info about bin_trips is in similarity.py (delete_bins)
+# For mapping the labels, in one cluster, we do have trips with user labels.
+# There are only two situations - trips are grouped in one cluster or the single trip in one cluster.
+# Ideally, we should have labels [11,11,11] for trips in one cluster after the second round.
+# In such a cluster, we should already have 1 or 2 trips that have user labels, the third one is predicted.
+# Then we label it with the same user labels as the previous two.
+# We don't have 1.0 homogeneity score for most of the cluster. It can also be [11,11,12].
+# Then for the third one, we will still label it with the same user labels as the previous two
+# (since they are in one cluster).
```
shankari, line 28: I think you need to fix this example. Even with `[11,11,12]`, the homogeneity score will be 1.0 because

> Non-perfect labelings that further split classes into more clusters can be perfectly homogeneous:

Also, "we don't have 1.0 homogeneity score" is an artifact of the data, not of the algorithm.

shankari: In fact, this makes it seem that we will always have a 1.0 homogeneity score because we will have `[11,11,11]` or [11,11,12]`

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,58 @@
...
+# in bins above cutoff
+# More info about bin_trips is in similarity.py (delete_bins)
+# For mapping the labels, in one cluster, we do have trips with user labels.
+# There are only two situations - trips are grouped in one cluster or the single trip in one cluster.
+# Ideally, we should have labels [11,11,11] for trips in one cluster after the second round.
+# In such a cluster, we should already have 1 or 2 trips that have user labels, the third one is predicted.
+# Then we label it with the same user labels as the previous two.
+# We don't have 1.0 homogeneity score for most of the cluster. It can also be [11,11,12].
+# Then for the third one, we will still label it with the same user labels as the previous two
+# (since they are in one cluster).
```
shankari, line 28:
```suggestion
# The homogeneity score reflects the degree to which a cluster consists only of trips with similar ground truthed labels.
# The labels can be drawn from different sets as long as the mapping is unique (e.g. ["A", "A", "C"] matches perfectly
# with [0,0,1]). Similar ground truthed trips can end up in different clusters and still get a perfect score;
# the metric only checks whether a cluster is homogenous, not whether it is complete. Concretely, ["A", "A", "A", "C"] matches perfectly with [0,0,1,2] because although the trips for A are split between clusters 0 and 1, each cluster only has trips labeled A.
# For our use case, this is fine because... please fill in here.
# We use the request percentage metric to check the completeness of the clusters.
```

tour_model_eval/get_scores.py
shankari: @corinne-hcr Added suggested text below; please fill in and commit.

tour_model_eval/get_scores.py
```diff
@@ -18,14 +18,12 @@ def compare_trip_orders(bins,bin_trips,filter_trips):
...
-# In such a cluster, we should already have 1 or 2 trips that have user labels, the third one is predicted.
-# Then we label it with the same user labels as the previous two.
-# We don't have 1.0 homogeneity score for most of the cluster. It can also be [11,11,12].
-# Then for the third one, we will still label it with the same user labels as the previous two
-# (since they are in one cluster).
+# The homogeneity score reflects the degree to which a cluster consists only of trips with similar ground truthed labels.
+# The labels can be drawn from different sets as long as the mapping is unique (e.g. ["A", "A", "C"] matches perfectly
+# with [0,0,1]). Similar ground truthed trips can end up in different clusters and still get a perfect score;
+# the metric only checks whether a cluster is homogenous, not whether it is complete. Concretely, ["A", "A", "A", "C"] matches perfectly with [0,0,1,2] because although the trips for A are split between clusters 0 and 1, each cluster only has trips labeled A.
+# For our use case, this is fine because... please fill in here.
```
shankari, line 25: @corinne-hcr "please fill in here"

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,56 @@
...
+# This function is to get homogeneity score after the first/second round of clustering
+# It is based on bin_trips, which are common trips. bin_trips are collected according to the indices of the trips
+# in bins above cutoff
+# More info about bin_trips is in similarity.py (delete_bins)
+# The homogeneity score reflects the degree to which a cluster consists only of trips with similar ground truthed labels.
+# The labels can be drawn from different sets as long as the mapping is unique (e.g. ["A", "A", "C"] matches perfectly
+# with [0,0,1]). Similar ground truthed trips can end up in different clusters and still get a perfect score;
+# the metric only checks whether a cluster is homogenous, not whether it is complete. Concretely, ["A", "A", "A", "C"] matches perfectly with [0,0,1,2] because although the trips for A are split between clusters 0 and 1, each cluster only has trips labeled A.
+# For our use case, this is fine because... please fill in here.
+# We use the request percentage metric to check the completeness of the clusters.
```
shankari, line 26:
```suggestion
# The labels can be drawn from different sets as long as the mapping is unique (e.g. ["A", "A", "C"] matches perfectly
# with [0,0,1]). 
# Ideally, there would be 1:1 mapping between labels and clusters - e.g. ["A", "A", "A"] maps to [1,1,1]
# This can break in two ways:
# label A maps to two different clusters - e.g. ["A", "A", "A"] maps to [1,2,3]. In this case, the homogeneity score will still be 1.0, since each cluster only has label "A". We capture this difference through the request percentage metric, which will result in three queries for [1,2,3] and only one for [1,1,1]
# two different labels map to the same cluster - e.g. ["A", "A", "B"] maps to [1,1,1]. This is the case captured by the homogeneity score, which will be .... please fill in here. This maps well to our use case because in this case... add short (1 sentence) explanation of bad labeling here.

Similar ground truthed trips can end up in different clusters and still get a perfect score;
# the metric only checks whether a cluster is homogenous, not whether it is complete. Concretely, ["A", "A", "A", "C"] matches perfectly with [0,0,1,2] because although the trips for A are split between clusters 0 and 1, each cluster only has trips labeled A.
# For our use case, this is fine because... please fill in here.
# We use the request percentage metric to check the completeness of the clusters.
```

tour_model_eval/get_scores.py
```diff
@@ -0,0 +1,66 @@
...
+# In this case, the homogeneity score will still be 1.0, since each cluster only has label "A".
+# For our use case, this is fine because the number of clusters depends on the Euclidean distance of the feature points,
+# which are extracted from trips, in multi-dimension. If trips with same user labels are far away from each other in the
+# multi-dimension, they will be separated in different clusters according to to cutoff. They could be separated in three
+# clusters. The reason is that they might have very different trajectory distance or duration.
+# We capture this difference through the request percentage metric, which will result in three queries for [1,2,3]
+# and only one for [1,1,1]
+# two different labels map to the same cluster - e.g. ["A", "A", "B"] maps to [1,1,1]. This is the case captured by the
+# homogeneity score, which will be less than 1.0 (0 representes inhomogeneous, 1.0 represents homogeneous).
+# This maps well to our use case because in this case the trips might have similar feature points.
```
shankari, line 36:
```suggestion
# For problem, this would map to the use case where trips with same user labels are actually to different 
# destinations. For `medical` or `personal` locations, for example, users could actually go to multiple medical 
# facilities or friends' houses. In this case, the trips will be in different clusters, but since the destinations are in 
# fact different, this would actually be the correct behavior.
# The trips could also be to the same location, but be clustered differently due to minor variations in duration or 
# distance (maybe due to traffic conditions). This could result in multiple clusters for what is essentially the same 
# trip. We capture this difference through the request percentage metric, which will result in three queries for 
# [1,2,3] and only one for [1,1,1]
# two different labels map to the same cluster - e.g. ["A", "A", "B"] maps to [1,1,1]. This is the case captured by the
# homogeneity score, which will be less than 1.0 (0 representes inhomogeneous, 1.0 represents homogeneous).
# This maps well to our use case because in this case, assigning the same label to all trips in the cluster would
# be incorrect. In particular, if we did not have the ground truth, the third trip would be labeled "A", 
# which would lower the accuracy.
```

PR #23, opened 2021-06-02
check in notebook code

corinne-hcr:


shankari:
Wait, is the entire notebook implementation in one ~ 300 line cell? You are right, I don't like the structure.

shankari:
@corinne-hcr I am loathe to ask for significant restructuring. changes given the imminent deadline, but this definitely needs to be restructured significantly after the paper is submitted. I will mark my review with [NOW] and [LATER] to indicate priority.

tour_model_eval/first_second_round_evaluation.ipynb
```diff
@@ -0,0 +1,352 @@
...
+    "                point_features.append([start[0], start[1], end[0], end[1], distance, duration])\n",
+    "\n",
+    "            x = np.array(point_features)\n",
+    "\n",
+    "            method = 'single'\n",
+    "            # get labels after two rounds of clustering on common trips\n",
+    "            new_labels = lp.get_new_labels(x, low, dist_pct, second_round_idx_labels, new_labels, method=method)           \n",
+    "            track = lp.change_track_labels(track, new_labels)\n",
+    "        \n",
+    "\n",
```
shankari, line 291: this is essentially the same as the tuning set. Why are you copy-pasting the code?

tour_model_eval/first_second_round_evaluation.ipynb
```diff
@@ -0,0 +1,352 @@
...
+    "# collect homogeneity score for the first/second round for all users\n",
+    "all_homogeneity_score_first_train = []\n",
+    "all_homogeneity_score_first_test = []\n",
+    "all_homogeneity_score_second_train = []\n",
+    "all_homogeneity_score_second_test = []\n",
+    "all_median_homo_first = []\n",
+    "\n",
+    "for a in range(len(all_users)):\n",
+    "    user = all_users[a]\n",
+    "    filter_trips, trips = preprocess.filter_data(user, radius)\n",
```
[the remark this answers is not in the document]
shankari: My original feedback was https://github.com/e-mission/e-mission-eval-private-data/pull/20/commits/bfdf1e8a77446e9b0364514382a2529832a8b752#r631490113

tour_model_eval/first_second_round_evaluation.ipynb
```diff
@@ -0,0 +1,352 @@
...
+    "                    for trip in second_round_trips:\n",
+    "                        start = trip.data.start_loc[\"coordinates\"]\n",
+    "                        end = trip.data.end_loc[\"coordinates\"]\n",
+    "                        distance = trip.data.distance\n",
+    "                        duration = trip.data.duration\n",
+    "                        point_features.append([start[0], start[1], end[0], end[1], distance, duration])\n",
+    "\n",
+    "                    x = np.array(point_features)\n",
+    "\n",
+    "                    method = 'single'\n",
```
shankari, line 181: justify usage of this method?

shankari (commented):
The most egregious part of this code is the duplication between tuning and testing. It is not too terrible otherwise. Still debating whether to merge or not.

tour_model_eval/first_second_round_evaluation.ipynb
```diff
@@ -0,0 +1,352 @@
...
+    "                    track = lp.change_track_labels(track, new_labels)\n",
+    "\n",
+    "                # get request percentage for the subset for the second round\n",
+    "                percentge_second = float('%.3f' % grp.get_req_pct(new_labels, track, filter_trips, sim))\n",
+    "\n",
+    "                # get homogeneity score for the second round\n",
+    "                homo_second = float('%.3f' % gs.score(bin_trips, new_labels))\n",
+    "\n",
+    "                curr_score = 0.5 * homo_second + 0.5 * (1 - percentge_second)\n",
+    "                curr_score = float('%.3f' % curr_score)\n",
```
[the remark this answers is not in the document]
shankari: the file could have multiple different `score` implementations

shankari (changes requested):
overall, I would recommend, at a high level, something like:

```
def tune(...):
   first_round
   second_round
   return tuning_parameters
```

```
def test(...., tuning_parameters):
    first_round
    second_round
    return results

File: tour_model_eval/first_second_round_evaluation.ipynb [hunk 5]
```diff
@@ -46,273 +57,81 @@
...
-    "    pct_collect_first_train = []\n",
-    "    # collect homogeneity score for a user for the first round\n",
-    "    homo_collect_first_train = []\n",
-    "    # collect request percentage for a user for the second round\n",
-    "    pct_collect_second_train = []\n",
-    "    # collect homogeneity score for a user for the second round\n",
-    "    homo_collect_second_train = []\n",
+    "    tune_data = preprocess.get_subdata(filter_trips, test_idx)\n",
+    "    test_data = preprocess.get_subdata(filter_trips, tune_idx)\n",
     "\n",
```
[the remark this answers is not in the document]
shankari: I would like the pipeline to see separate steps like:

```
training_trips = get_subset(....)
training_features = extract_features(training_trips)
```

tour_model_eval/first_second_round_evaluation.ipynb
```diff
@@ -46,273 +57,81 @@
...
+    "    coll_tune_score = []\n",
+    "    coll_tradeoffs = []\n",
+    "    # tuning\n",
+    "    pct_collect_first,homo_collect_first,pct_collect_second,homo_collect_second,coll_tradeoffs,  coll_tune_score= ep.tuning_test(tune_data,radius,pct_collect_first,homo_collect_first,pct_collect_second,homo_collect_second,coll_tradeoffs,coll_tune_score,tune = True)\n",
     "\n",
-    "                # get homogeneity score for the second round\n",
-    "                homo_second = float('%.3f' % gs.score(bin_trips, new_labels))\n",
+    "    pct_collect_first, homo_collect_first, pct_collect_second, homo_collect_second = ep.init_score()\n",
+    "    # testing\n",
+    "    pct_collect_first,homo_collect_first,pct_collect_second,homo_collect_second,coll_tradeoffs,coll_tune_score = ep.tuning_test(test_data,radius, pct_collect_first, homo_collect_first,pct_collect_second,homo_collect_second, coll_tradeoffs,coll_tune_score,test=True)\n",
```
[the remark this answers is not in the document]
corinne-hcr: At this point, I am not sure if we still need the tradeoffs after the clustering, so I keep it.
I am still figuring out how to predict the labels

corinne-hcr:
I have separated the functions from a giant function, basically like tune(...) and test(...). but not sure if it is ok now. I plan to move the `evaluation_pipeline.py` to e-mission-server

tour_model_eval/evaluation_pipeline.py
```diff
@@ -0,0 +1,173 @@
+import emission.analysis.modelling.tour_model.similarity as similarity
+import numpy as np
+import emission.analysis.modelling.tour_model.get_request_percentage as grp
+import emission.analysis.modelling.tour_model.get_scores as gs
+import emission.analysis.modelling.tour_model.label_processing as lp
+import emission.analysis.modelling.tour_model.data_preprocessing as preprocess
+
+def second_round(first_label_set,first_labels,bin_trips,filter_trips,low,dist_pct,sim,new_labels,track):
```
shankari, line 8: do you really need 9 mandatory arguments to this function? What do you need `sim` for, for example, if you already have `filter_trips` and `track`?

do you really need both `first_label_set` and `first_labels`? Why can you not get `first_label_set` from `first_label` in this function?

At a high level, if you get to so many arguments, I will recommend that you create a class to pass them in. But given our time constraints, I am not going to push for this now.

tour_model_eval/evaluation_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+    # bins contain common trips indices
+    if len(bins) is not 0:
+        gs.compare_trip_orders(bins, bin_trips, filter_trips)
+        first_labels = get_first_label(bins)
+        # new_labels temporary stores the labels from the first round, but later the labels in new_labels will be
+        # updated with the labels after two rounds of clustering.
+        new_labels = first_labels.copy()
+        first_label_set = list(set(first_labels))
+        track = get_track(bins, first_labels)
+        # collect tuning scores and parameters
```
shankari, line 92: all of this is common, so it seems like it can be moved into `first_round` or into a new function like `get_first_round_labels` or something which is called from both `tune` and `test`

tour_model_eval/evaluation_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+                percentage_second, homo_second = second_round(first_label_set, first_labels, bin_trips, filter_trips,
+                                                              low, dist_pct,
+                                                              sim, new_labels, track)
+
+                curr_score = gs.get_score(homo_second, percentage_second)
+                if curr_score not in tune_score:
+                    tune_score[curr_score] = (low, dist_pct, homo_second, percentage_second)
+
+        best_score = max(tune_score)
+        sel_tradeoffs = tune_score[best_score][0:2]
```
shankari, line 105: again, it seems like it would be good to create a class called ClusteringResults with these fields or even return a simple dict with the fields instead of passing around lists that are hard to read. Lists are convenient for you to write, but code like `[0:2]` doesn't indicate to others what you are returning.

tour_model_eval/evaluation_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+    tune_idx, test_idx = preprocess.split_data(filter_trips)
+    tune_data = preprocess.get_subdata(filter_trips, test_idx)
+    test_data = preprocess.get_subdata(filter_trips, tune_idx)
+    pct_collect_first = []
+    homo_collect_first = []
+    pct_collect_second = []
+    homo_collect_second = []
+    coll_score = []
+    coll_tradeoffs = []
+
```
shankari, line 153: Again, can this be moved into a list of results or even a dataframe instead of multiple lists of results?

tour_model_eval/evaluation_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+    for k in range(len(test_data)):
+        tradoffs = coll_tradeoffs[k]
+        low = tradoffs[0]
+        dist_pct = tradoffs[1]
+        homo_first, percentage_first, homo_second, percentage_second, scores = test(tune_data[k],radius,low,dist_pct)
+        pct_collect_first.append(percentage_first)
+        homo_collect_first.append(homo_first)
+        pct_collect_second.append(percentage_second)
+        homo_collect_second.append(homo_second)
+        coll_score.append(scores)
```
shankari, line 169: then this will be significantly simplified.

shankari (changes requested):
Almost there!

tour_model_eval/load_predict.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+
+if __name__ == '__main__':
+    participant_uuid_obj = list(edb.get_profile_db().find({"install_group": "participant"}, {"user_id": 1, "_id": 0}))
+    all_users = [u["user_id"] for u in participant_uuid_obj]
+    user = all_users[0]
+    radius = 100
+    trips = preprocess.read_data(user)
+    filter_trips = preprocess.filter_data(trips, radius)
```
shankari, line 90: now you see how it helps to have these functions pulled out into a separate module.

tour_model_eval/load_predict.py
```diff
@@ -0,0 +1,92 @@
+import emission.core.get_database as edb
+import emission.analysis.modelling.tour_model.similarity as similarity
+import emission.analysis.modelling.tour_model.data_preprocessing as preprocess
+import jsonpickle as jpickle
+
+
+def loadModelStage(filename):
+    import emission.analysis.classification.inference.mode.seed.pipeline as seedp
+    model = loadModel(filename)
+    return model
```
shankari, line 10: why do you need this import?
In general, you should not use anything from the mode inference pipeline, since that has separate functionality. It is just an example.

tour_model_eval/evaluation_pipeline.py
corinne-hcr: `sim` is to find out trips below the cutoff in the 1st round. Those trips didn't go through further analysis except the 1st round, so I have to use `sim`

> At a high level, if you get to so many arguments, I will recommend that you create a class to pass them in. But given our time constraints, I am not going to push for this now.

I don't understand this. Maybe I can reduce 1 or 2 arguments. But given that there are still many arguments I have to pass in, what is the difference between writing a pure function to accept them and writing a class?

tour_model_eval/evaluation_pipeline.py
```diff
@@ -0,0 +1,173 @@
+import emission.analysis.modelling.tour_model.similarity as similarity
+import numpy as np
+import emission.analysis.modelling.tour_model.get_request_percentage as grp
+import emission.analysis.modelling.tour_model.get_scores as gs
```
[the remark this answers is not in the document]
corinne-hcr: files added

tour_model_eval/load_predict.py
```diff
@@ -0,0 +1,92 @@
...
+    sel_fl = None
+    for fl in first_round_label_set:
+        # extract location features of selected bin
+        sel_loc_feat = bin_locations[fl]
+        # Check if start/end locations of the new trip and every start/end locations in this bin are within the range of
+        # radius. If so, the new trip falls in this bin. Then predict the second round label of the new trip
+        # using this bin's model
+        if in_bin(sel_loc_feat, trip_loc_feat, radius):
+            sel_fl = fl
+            break
```
shankari, line 65: This is too intensive for prediction. Basically, for prediction, you do **not** want to write any code that iterates over all the trips for a user; it will not scale.

It will not scale means that if you have a user with data for 6 months, the server will crash or it will take hours for the computation.

tour_model_eval/load_predict.py
```diff
@@ -0,0 +1,92 @@
...
+    user = trip[0]['user_id']
+    trip_feat = preprocess.extract_features(trip)[0]
+    trip_loc_feat = trip_feat[0:4]
+    try:
+        # load locations of bins(1st round of clustering)
+        bin_locations = loadModelStage('locations_' + str(user))[0]
+        # load models from the 2nd round of clustering
+        models = loadModelStage('models_' + str(user))[0]
+        # load user labels in all clusters
+        user_labels = loadModelStage('user_labels_' + str(user))[0]
```
shankari, line 51: can you add some information to the PR or the commit about what these data structures look like?

tour_model_eval/load_predict.py
```diff
@@ -0,0 +1,92 @@
...
+        # using this bin's model
+        if in_bin(sel_loc_feat, trip_loc_feat, radius):
+            sel_fl = fl
+            break
+    if not sel_fl:
+        return {}
+    # choose selected model
+    sel_model = models[sel_fl]
+    # predict 2nd label for the new trip
+    sel_sl = str(sel_model.predict([trip_feat])[0])
```
shankari, line 71: why do you need to get the zeroth element here?

tour_model_eval/load_predict.py
```diff
@@ -0,0 +1,92 @@
...
+    # values of the selected 1st round label
+    sel_1st_round_val = user_labels[sel_fl][0]
+    second_label_ls = list(sel_1st_round_val.keys())
+    if sel_sl not in second_label_ls:
+        return {}
+    # values of the selected 2nd round label, wrapped in a list
+    sel_2nd_round_val = sel_1st_round_val[sel_sl]
```
shankari, line 78: can you elaborate a bit more on what these do?

shankari (changes requested):
- what is the structure of the `user_labels` and `models` data structures?
- reading all user trips is too intensive
- don't understand why `predict` is not good enough to directly find the labels and has to go through first round and second round again

tour_model_eval/load_predict.py
corinne-hcr: I didn't actually run all trips. I saved all coordinates for all common trips in the `build and save` step. My logic here is to compare the new trip's coordinates with other trips'  to see which bin it should fall in. 
The other way I can think of is that finding a representative trip for every bin, then compare it with the new trip. But there is a risk that new trip match a representative trip but not all the existing trips in that bin. ( I thought that the new trips would match all trips in a bin if it matched one, but that would not be certain.)
Do you have other way to see if a new trip can fall in any bin?

tour_model_eval/load_predict.py
corinne-hcr: If I remove this, there will be an error.
```
AttributeError: attribute 'dtype' of 'numpy.generic' objects is not writable
```
But I figure that you have these two line in your pipeline file. So I put them in my function. It works now.
```
    import jsonpickle.ext.numpy as jsonpickle_numpy
    jsonpickle_numpy.register_handlers()
```

tour_model_eval/load_predict.py
shankari: There is no concept of running a trip. A trip is data, you can only run code.
I think that you mean that you are not running the clustering algorithm on all trips.
You are not running the clustering algorithm, but you are still running a `for` loop over all the trips aka "iterating over the trips". Iterating over 10k trips to check the distances is still going to take time.

Using a representative trip has the problem that you have outlined " I thought that the new trips would match all trips in a bin if it matched one, but that would not be certain". A better option seems to be to store a geometric shape representing all the starts and all the ends of all the trips. You can then quickly check whether the new trip is within the shape. 

The simplest shape is probably a circle. A potential algorithm might be:
- get all the points
- find the centroid
- find the max distance of any point from the centroid
- use that as the circle representing the points

You can also use a minimum enclosing circle algorithm/implementation instead of the hacky solution above.
https://www.nayuki.io/page/smallest-enclosing-circle

shankari: Actually, you can't use that implementation because its license is LGPL and we don't want to taint our BSD-3 license.
An alternative is OpenCV, which uses an Apache license.

shankari: A lighter weight alternative appears to be
https://github.com/luwei14/ShapeAnalyzer

Make sure to update the `ThirdPartyLicenses.md` file if you include a new library

tour_model_eval/load_predict.py
shankari: is there a new change showing this so I can resolve this conversation?

tour_model_eval/load_predict.py
shankari: This has a super simple example using scipy
https://www.cs.au.dk/~gerth/ipsa18/slides/optimization.pdf

tour_model_eval/load_predict.py
corinne-hcr: added comments and examples, also changed the variable name

tour_model_eval/load_predict.py
corinne-hcr: added explanation

tour_model_eval/load_predict.py
corinne-hcr: added

tour_model_eval/evaluation_pipeline.py
shankari: you can write a class/struct to encapsulate them so that people don't have to write really long lines passing them back and forth. You can have a single class that you both pass in **and return** and that has the current state of the algorithm. Makes everything a lot less verbose.

tour_model_eval/load_predict.py
```diff
@@ -44,13 +45,29 @@ def predict_labels(trip):
     trip_loc_feat = trip_feat[0:4]
     try:
         # load locations of bins(1st round of clustering)
+        # e.g.{'0': [[start lon1, start lat1, end lon1, end lat1],[start lon, start lat, end lon, end lat]]}
+        # another explanation: -'0': label from the 1st round
+        #                      - the value of key '0': all trips that in this bin
+        #                      - for every trip: the coordinates of start/end locations
         bin_locations = loadModelStage('locations_' + str(user))[0]
+
```
shankari: this is going to be completely changed in the next commit, so ignoring for now.

tour_model_eval/load_predict.py
```diff
@@ -44,13 +45,29 @@ def predict_labels(trip):
...
+        # the following is an example of the saved models.
+        # e.g. {'0': KMeans(n_clusters=2, random_state=0), '1': KMeans(n_clusters=5, random_state=0)}
         models = loadModelStage('models_' + str(user))[0]
+
         # load user labels in all clusters
+        # assume that we have 1 cluster(bin) from the 1st round of clustering, which has label '0',
+        # and we have 1 cluster from the 2nd round, which has label '1'
+        # the value of key '0' contains all 2nd round clusters
+        # the value of key '1' contains all user labels and probabilities in this cluster
+        # e.g. {'0': [{'1': [{'labels': {'mode_confirm': 'shared_ride', 'purpose_confirm': 'home', 'replaced_mode': 'drove_alone'}}]}]}
```
shankari, line 66: is there a reason why this is not just

```
{'01': [{'labels': {'mode_confirm': 'shared_ride', 'purpose_confirm': 'home', 'replaced_mode': 'drove_alone'}}]}]}
```

Also, I want to confirm that you are in fact storing multiple possible labels and the associated probabilities, because that is not what I see here.

I would expect this to actually be

```
[{'labels': {'mode_confirm': 'shared_ride', 'purpose_confirm': 'home', 'replaced_mode': 'drove_alone'}, "p": 1.0}]
```

tour_model_eval/load_predict.py
```diff
@@ -64,18 +81,28 @@ def predict_labels(trip):
...
+    # '1': [{'labels': {'mode_confirm': 'shared_ride', 'purpose_confirm': 'church', 'replaced_mode': 'drove_alone'},
+    # 'p': 0.9333333333333333},{'labels': {'mode_confirm': 'shared_ride', 'purpose_confirm': 'entertainment',
+    # 'replaced_mode': 'drove_alone'},'p': 0.06666666666666667}]}
+    # more explanation: '0' and '1' are 2nd round clusters from a bin(1st round cluster)
+    # cluster '0' contains all user label combinations and probabilities in this cluster
+    seccond_round_result = user_labels[sel_fl][0]
+    second_label_ls = list(seccond_round_result.keys())
     if sel_sl not in second_label_ls:
-        return {}
+        return []
```
shankari: why do you need this? I think you can directly use `sel_ls not in second_round_result`

tour_model_eval/load_predict.py
```diff
@@ -84,9 +111,36 @@ def predict_labels(trip):
...
+    radius = 100
+    trips = preprocess.read_data(user)
+    filter_trips = preprocess.filter_data(trips, radius)
+    new_trip = [filter_trips[0]]
+    # result is []
+    print(predict_labels(new_trip))
+
+    # case 4: the new trip falls in a 1st round bin, but predict to be a new cluster in the 2nd round
+    # result is []
+    # no example for now
```
shankari: Glad to see some testing done, even if it is not fully complete and is not in the unit test framework.

tour_model_eval/load_predict.py
corinne-hcr: Because Gabriel's function need a list, I follow the result from `placeholder_predictor_2(trip)`

shankari: this is not about the return type. It is about the need for `second_label_ls = list(seccond_round_result.keys())`

tour_model_eval/load_predict.py
corinne-hcr: I have two rounds of clustering. Let's say the final result has 2 layers. The first layer is result from the first round, the second layer is the result from the second round. So, in the first layer, every bin contains all second round labels and associate user labels and `p` in each second round cluster

What you expect is my final result. When the new trip matches a 2nd round cluster, then the associate labels and `p` would look like that.

tour_model_eval/load_predict.py
corinne-hcr: Oh ok, I see your point.

tour_model_eval/load_predict.py
shankari: > I have two rounds of clustering. Let's say the final result has 2 layers. The first layer is result from the first round, the second layer is the result from the second round. 

I understand but you are not addressing my comment. Why is it not enough to have the label be the concatenation of the first and second rounds - e.g. `{'0': {'1': ...}` to `{'01': {....`

corinne-hcr: I don't have the model that labeled '1', '2', '11', '12'... since the first round was not from any library. What Kmeans did was label the 2nd round cluster [0,1,2,3,...]. What I got using Kmeans `predict` is like [1]. Therefore, it is better to separate into 2 layers instead of getting concatenation of labels. 
I have to get the 1st label anyway, then `predict` gives me the 2nd label. I just use the 2nd label to get the result. Otherwise, in the labels saving step, it would be complicated. Also, in the predict step, I have to concatenate the labels again.

shankari: None of the reasons make sense to me ("otherwise in the labels saving step, it would be complicated" Why?), but it is also not worth arguing further about this.

Although you are not able to justify it, your design decision is not a-priori incorrect, so I'm going to move on.

shankari (changes requested):
I don't see any new design problems here.
- You still need to fix the `locations` model.
- The `main` function is way too large (in general, it is good to have functions be ~ 1 page long. But I will not force this now
- you tend to treat dataframes as lists and iterate over them. using vectorized operations is more efficient - e.g. https://towardsdatascience.com/efficient-pandas-apply-vs-vectorized-operations-91ca17669e84

tour_model_eval/build_save_model.py
```diff
@@ -0,0 +1,170 @@
...
+                    one_set_labels = {}
+                    for i in range(len(unique_labels)):
+                        # e.g. labels_only={'mode_confirm': 'pilot_ebike', 'purpose_confirm': 'work', 'replaced_mode': 'walk'}
+                        labels_only = {column: unique_labels.iloc[i][column] for column in labels_columns}
+                        one_set_labels["labels"] = labels_only
+                        one_set_labels['p'] = unique_labels.iloc[i]['p']
+                        # e.g. one_set_labels = {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 1.0}
+                        # in case append() method changes the dict, we use deepcopy here
+                        labels_set = copy.deepcopy(one_set_labels)
+                        sec_label_obj.append(labels_set)
```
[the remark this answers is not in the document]
corinne-hcr: line 126 `append()`
This will change the dict if the dict is not an entire new object. if I directly append `one_set_labels`, the values in `one_set_labels` will be changed.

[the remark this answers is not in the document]
shankari: really? because that is not how python dicts or lists work. see below.
If `one_set_labels` is changed, there is some other reason, you need to understand and explain it.

```
>>> sec_label_obj = []
>>> one_set_labels = {}
>>> one_set_labels["labels"] = {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}
>>> one_set_labels["p"] = 1.0
>>> sec_label_obj.append(one_set_labels)
>>> sec_label_obj
[{'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 1.0}]
>>> one_set_labels
{'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 1.0}
```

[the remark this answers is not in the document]
corinne-hcr: Here you are using one dict. But as you append more dicts with same name(`one_set_labels`), the previous added dict will be changed.
```
>>> test_list = []
>>> a={'q':1,'w':[]}
>>> test_list.append(a)
>>> print(test_list)
[{'q': 1, 'w': []}]

>>> a['q']=2
>>> a['w'].append(123)
>>> test_list.append(a)
>>> print(test_list)
[{'q': 2, 'w': [123]}, {'q': 2, 'w': [123]}]
```

Several days ago, I found some explanations in Chinese. Someone says that what appended to the list is the address of the dict. It doesn't create a new object in the list. So, every time we append a dict with the same name, it actually points to the same address. 

Also, there is a difference between `copy()` and `deepcopy()`
`copy()` is shallow copy
```
>>> a={'q':1,'w':[]}
>>> b=a.copy()
>>> b['q']=2
>>> b['w'].append(123)
>>> print(a)
>>> print(b)
{'q': 1, 'w': [123]}
{'q': 2, 'w': [123]}
```
If we change to `deepcopy()`, we can have entirely new object.
```
>>> a={'q':1,'w':[]}
>>> b=copy.deepcopy(a)
>>> b['q']=2
>>> b['w'].append(123)
>>> print(a)
>>> print(b)
{'q': 1, 'w': []}
{'q': 2, 'w': [123]}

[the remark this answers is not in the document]
Comment (shankari) [reply tour_model_eval/build_save_model.py:126]: Here you are using one dict. But as you append more dicts with same name(one_set_labels), the previous added dict will be changed. This is not true if you do it "pythonically"

```
>>> sec_label_obj = []
>>> for p in range(0,10,1):
...     one_set_labels = {}
...     one_set_labels["labels"] = {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}
...     one_set_labels["p"] = p/10
...     sec_label_obj.append(one_set_labels)
...
>>> sec_label_obj
[{'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.0}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.1}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.2}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.3}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.4}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.5}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.6}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.7}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.8}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.9}]
```

This style is more pythonic because it lends itself to list comprehensions, use with itertools, functional style programming, etc

```
>>> sec_label_obj = [{"labels": {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, "p": p/10} for p in range(0,10,1)]
>>> sec_label_obj
[{'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.0}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.1}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.2}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.3}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.4}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.5}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.6}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.7}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.8}, {'labels': {'mode_confirm': 'walk', 'replaced_mode': 'walk', 'purpose_confirm': 'exercise'}, 'p': 0.9}]
```

> Several days ago, I found some explanations in Chinese. Someone says that what appended to the list is the address of the dict. It doesn't create a new object in the list. So, every time we append a dict with the same name, it actually points to the same address.

@corinne-hcr I am aware of the difference between shallow copy and deepcopy. But the difference is primarily when you **modify** objects that you have appended to the list (e.g. `b['q']=2`); merely appending the object doesn't change it. In general modifying objects that you are appending to the list is bad practice because the code cannot be easily refactored using the alternate styles I outlined above.

Where are you modifying objects in your code?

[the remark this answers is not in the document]
Comment (corinne-hcr) [reply tour_model_eval/build_save_model.py:126]: I think I understand the differences between your output and mine. I created `one_set_labels = {}` before the loop, then I didn't have a new object every time I change the values. You created `one_set_labels = {}` in the loop, so you have a new object every time. But I am going to change the structure. Glad to figure it out!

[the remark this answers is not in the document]
Comment (shankari) [reply tour_model_eval/build_save_model.py:126]: Creating one object outside the loop and then deepcopying in the loop is not pythonic. It will make it harder to maintain the code in the future.

**Comment by shankari** [comment]
@corinne-hcr just to clarify, I don't plan to merge this PR at this point, you can continue working on the refactoring and let me know when you are completely done. I will copy the modeling code to the server repo and will be working there directly.

I will submit a separate PR there as well. I don't think you run the modeling code from the notebook, but if you do, I can just write a harness anyway.

**Comment by corinne-hcr** [comment]
I don't think the notebook call the modeling code. But the notebook generate results and parameters. The modeling code need the parameters.
So, should I keep refactoring the modeling code and let you know when I am done?

**Comment by shankari** [comment]
@corinne-hcr I have already started working on moving the modeling code over. You are welcome to refactor the modeling code in parallel if you like, but I cannot wait until you are done.

The modelling code should generate the results and parameters internally, there cannot be a notebook sitting around when it is running on production.

**Comment by shankari** [comment]
you may want to move over the `.py` files into the server repo yourself first so you get appropriate credit for them.
If I copy them over and commit them, it will look like I was the one that wrote all the code.

**Comment by corinne-hcr** [comment]
But if the modeling code need to generate the best parameters, it has to run through the same thing in the notebook code. It takes a long time, and it is repeated. 
I thought that all results were saved after running the notebook code. Then run the modeling code to find the best result(best split, parameters) and re-build & save the models

**Comment by shankari** [comment]
@corinne-hcr is there an ETA for copying the modeling files over? otherwise, I am just going to do it myself.

**Comment by corinne-hcr** [comment]
> @corinne-hcr is there an ETA for copying the modeling files over? otherwise, I am just going to do it myself.

Do you mean moving everything to the server repo except the notebook code?

**Comment by shankari** [comment]
@corinne-hcr 

> Do you mean moving everything to the server repo except the notebook code?

Yes. There are also some differences between the same files in the server repo and this repo; we need to have the most recent version (whichever that is) in the server repo - e.g.

```
--- ./get_users.py  2021-07-19 11:51:10.000000000 -0700
+++ /Users/kshankar/e-mission/e-mission-eval-private-data/tour_model_eval/./get_users.py    2021-05-27 17:39:27.000000000 -0700
@@ -1,4 +1,4 @@
-import emission.analysis.modelling.tour_model.data_preprocessing as preprocess
+import data_preprocessing as preprocess


 # to determine if the user is valid:
@@ -19,8 +19,7 @@
     for i in range(len(all_users)):
         curr_user = 'user' + str(i + 1)
         user = all_users[i]
-        trips = preprocess.read_data(user)
-        filter_trips = preprocess.filter_data(trips,radius)
+        filter_trips,trips = preprocess.filter_data(user,radius)
         if valid_user(filter_trips,trips):
             valid_user_ls.append(curr_user)
             user_ls.append(curr_user)
```

**Comment by shankari** [comment]
> But if the modeling code need to generate the best parameters, it has to run through the same thing in the notebook code. It takes a long time, and it is repeated.

> I thought that all results were saved after running the notebook code. Then run the modeling code to find the best result(best split, parameters) and re-build & save the models

The notebook is primarily for visualization and interactive exploration. I think you are arguing that we should pick parameters separate (in the notebook) from building the clusters (in the model). That is a good idea; we might want to generate the optimal parameters only once every 6 months or something.

Three points:
- I don't see why picking the best parameters needs to happen in the notebook. There is no human intervention in picking the parameters, nothing that you need to "show" to a user.
- The parameters could be different for different regions and different demographic groups. They might even change over time for the same group.
- I am not going to sit there running the parameter generation code on my laptop for multiple different pilots.

Again, for this first implementation, tonight, we can even use hardcoded parameters that you got from the notebook. But it is not a long-term solution.

And moving the code to the server does not preclude importing it and running it in a notebook. So let's move things over to the server, please. Right now.

**Comment by shankari** [comment]
@corinne-hcr is there an ETA for resolving the differences between this repo and the server and moving the other .py files over? I can take a break for lunch to give you some time to work on this but I can't wait any longer. If I copy the files over, they will appear to have been created by me instead of you.

**Comment by corinne-hcr** [comment]
The codes has moved. But the PR is in progress of checks. 
If you don't want to run the notebook to generate the result, can I put the part of generating codes in my evaluation_pipeline.py in main()? It is just not calling the scatter plot function.

**Comment by shankari** [comment]
Not sure what this would look like, please submit commit showing your idea.

**Comment by corinne-hcr** [comment]
I have updated the notebook code here

File: tour_model_eval/build_save_model.py [hunk 24]
```diff
@@ -0,0 +1,170 @@
...
+def save_models(obj_name,obj,user):
+    obj_capsule = jpickle.dumps(obj)
+    filename = obj_name + '_' + str(user)
+    with open(filename, "w") as fd:
+        fd.write(obj_capsule)
+
+
+def main():
+    participant_uuid_obj = list(edb.get_profile_db().find({"install_group": "participant"}, {"user_id": 1, "_id": 0}))
+    all_users = [u["user_id"] for u in participant_uuid_obj]
```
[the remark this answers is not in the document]
shankari: I'm going to fix this.

tour_model_eval/build_save_model.py
```diff
@@ -0,0 +1,170 @@
...
+        if len(bins) is not 0:
+            gs.compare_trip_orders(bins, bin_trips, filter_trips)
+            first_labels = ep.get_first_label(bins)
+            first_label_set = list(set(first_labels))
+
+            # second round of clustering
+            model_coll = {}
+            bin_loc_feat = {}
+            fitst_round_labels = {}
+            for fl in first_label_set:
```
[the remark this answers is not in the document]
shankari: I guess I will fix this as well.

tour_model_eval/build_save_model.py
```diff
@@ -0,0 +1,170 @@
...
+                # get labels from the 2nd round of clustering
+                second_labels = kmeans.labels_
+
+                first_label_obj = []
+
+                # save user labels for every cluster
+                second_label_set = list(set(second_labels))
+                sec_round_labels = {}
+                for sl in second_label_set:
+                    sec_sel_trips = []
```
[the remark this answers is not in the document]
shankari: I guess I will fix this as well.

tour_model_eval/build_save_model.py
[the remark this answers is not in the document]
shankari: @corinne-hcr I still see the `deepcopy` here

tour_model_eval/first_second_round_evaluation.ipynb
```diff
@@ -0,0 +1,119 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "cathedral-pointer",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "participant_uuid_obj = list(edb.get_profile_db().find({\"install_group\": \"participant\"}, {\"user_id\": 1, \"_id\": 0}))\n",
+    "all_users = [u[\"user_id\"] for u in participant_uuid_obj]"
```
shankari, line 33: @corinne-hcr The participant UUID check was specific to the CanBikeCO mini-pilot - it is not true for other programs or datasets. I have created a new mongodump of only the participant data, you should load that instead and change this to the call from `emission.storage.timeseries.abstract_timeseries`

tour_model_eval/first_second_round_evaluation.ipynb
```diff
@@ -0,0 +1,119 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "following-fundamentals",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# get all/valid user list\n",
+    "user_ls, valid_users = gu.get_user_ls(all_users, radius)\n",
+    "# get all filenames of clustering result\n",
+    "collect_filename = predict.loadModelStage(\"collect_filename\")"
```
shankari, line 56: I don't understand why you need to have the filenames in a list. Can't you just read the file based on the UUID? Can you clarify your design decision here?

shankari:
@corinne-hcr I don't think you have addressed all of my comments. I am going to address most of them myself, but there are a few really small and clear one where it might be good for you to fix. Also what is the difference between the two notebooks? A short markdown summary would be useful.

tour_model_eval/first_second_round_evaluation.ipynb
corinne-hcr: Oh, I think this one can be deleted. I did use the UUID to save the files. The original design is to save the filename. But it seems not necessary now.

tour_model_eval/evaluation_pipeline.py
corinne-hcr: I have created a class for the second round. See `second_round_of_clustering.py`
Although it is not a perfect one, but it works.

tour_model_eval/evaluation_pipeline.py
corinne-hcr: There were just two lines of code repeated after refactoring, but I created a function         `get_first_label_and_track(bins,bin_trips,filter_trips)`, which can be called from both `tune` and `test`

tour_model_eval/evaluation_pipeline.py
corinne-hcr: I have already changed this part, there is no more`[0:2]`. I do have a dict to collect the parameters. The parameters are clearly pointed out and return separately. 
e.g.     return low,dist_pct

tour_model_eval/evaluation_pipeline.py
corinne-hcr: I have changed to use a dataframe to collect all clustering result. These lists are deleted.

tour_model_eval/evaluation_pipeline.py
corinne-hcr: These are deleted.

corinne-hcr:
Only `first_second_round_evaluation.ipynb` is the useful notebook. In the previous version, I went through the clustering in the notebook and saved the result as multiple lists. Now I change the result to one dataframe for one user, and the notebook is just to show the scatter plots for 2 rounds of clustering. The clustering code has been moved to evaluation_pipeline.

I think I have addressed your comments. Why do you see no difference in the notebook? Please see commit https://github.com/e-mission/e-mission-eval-private-data/pull/23/commits/bfd6ba6aad8cdfb89b8ae606e12ea024f0f06a54
But this I forgot to delete the output, so I commit a new one, which is latest version.

shankari:
> I think I have addressed your comments. 

I still see the `deepcopy` code, for example.

corinne-hcr:
That's in the `evaluation_pipeline.py`. Because that file is actually in `Modeling and functions #829` You said no more changes made in the file to avoid conflicts. I was trying to refactor it according to your suggestion. But that didn't return a result in right format for Gabriel's function. I thought you were going to refactor it by yourself. 
You might have a different version of evaluation_pipeline now. Should I delete the `collect filename` from there?

shankari:
Fine, I will refactor the code in the `.py` files myself. Once you explain the difference between the two .ipynb files, I will go ahead and merge this.

shankari:
@corinne-hcr is there at ETA for updating the two notebooks so I can merge this?

corinne-hcr:
I face a problem now. 
I delete the `collect_filename` and let the `plot` code read files from `user id`. When I test running, the notebook always shows that files not found. The `plot` file is in e-mission-server, the clustering result files are also in e-mission-server. The notebook is in the e-mission-eval-private-data repo. 
```
FileNotFoundError: [Errno 2] No such file or directory: 'user_576e37c7-ab7e-4c03-add7-02486bc3f42e.csv'
```
The file can be read by the `plot` code. I don't know why the notebook has this error. Previously I did everything in the e-mission-eval-private-data repo, so I didn't have such a problem.

shankari:
> The file can be read by the plot code. I don't know why the notebook has this error. Previously I did everything in the e-mission-eval-private-data repo, so I didn't have such a problem.

You know the reason why it has this error and have already identified it below.

> Previously I did everything in the e-mission-eval-private-data repo, so I didn't have such a problem.

You need to specify the path to the files so they can be read correctly.

corinne-hcr:
I think you can merge the code now.

shankari:
There is so much refactoring left for this code... but making the changes has now taken around two months.
This PR is way past its "sell by date", so merging and will deal with refactoring later.

corinne-hcr:
I delete the `collect_filename`in the notebook and in the plot code, since I will change my typo in evaluation_pipeline anyway, I can delete the `collect_filename`there.

PR #25, opened 2021-06-29
incomplete test code

corinne-hcr:
The tests need to get inputs. No values for comparisons at this point(I don't know what to compare right now).
Also need help with setup and teardown.


corinne-hcr:
moved to emission server repo

PR #31, opened 2022-06-07
Update README.md

allenmichael099:
Clarified what the eval_folder is and added windows versions of some commands

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,14 +9,28 @@ it needs to be included while running them.
 
 1. Set the home environment variable
 
+    Mac:
+
     ```
     $ export EMISSION_SERVER_HOME=<path_to_emission_server_repo>
     ```
+    
+    In windows command prompt use:
+    
+    ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,24 +9,42 @@ it needs to be included while running them.
...
+
     ```
     $ export EMISSION_SERVER_HOME=<path_to_emission_server_repo>
     ```
+    
+    In windows command prompt use:
+    
+    ```
+    > set EMISSION_SERVER_HOME=<path_to_emission_server_repo>
+    ```
```
shankari, line 22: Do we have to support windows? We are not testing the windows codepaths regularly, and I am concerned that this will just bitrot and then we have to debug it all over again. Can we at least indicate that windows is unsupported and people are using it at their own risk?

shankari (changes requested):

README.md
shankari: @allenmichael099 are you planning to resolve this so that I can merge it?

PR #36, opened 2023-07-31
Covering up a possible error : db.conf NOT FOUND

humbleOldSage:
Setting the repository up for the first time might cause this error to pop up. A solution was proposed by Shankari earlier in Teams chat . Just migrating it here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,12 +13,17 @@ it needs to be included while running them.
     $ export EMISSION_SERVER_HOME=<path_to_emission_server_repo>
     ```
 
+     To verify, check the environment variables using 
+
+        $ env
+
+    and ensure ```ENV_SERVER_HOME``` is present in the list and has the right path (as mentioned above).
+    
 1. If you haven't setup before, set up the evaluation system
 
     ```
     $ source setup.sh
     ```
- 
  1. If you have, activate
 
     ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,8 +18,8 @@ it needs to be included while running them.
     ```
     $ source setup.sh
     ```
-
-1. If you have, activate
+ 
+ 1. If you have, activate
```
shankari, line 22: Please don't include extraneous whitespace changes. It messes up the blame and makes everything harder to work with in the future.

README.md
```diff
@@ -38,6 +38,37 @@ $ cp -r ../conf .
...
+>    $ env
+>    ```
+>
+>
+>* Move the conf files AGAIN using :
+>
+>    ```console
+>    cp -r $EMISSION_SERVER_HOME/conf .
+>    ```
+>
```
shankari, line 68: This is not moving the files, it is copying them.

shankari (changes requested):
At a high level, I don't understand why you have to replicate instructions. There are already instructions to set the EMISSION_SERVER_HOME on line 12. Why not clarify those instructions instead of creating a whole new set of instructions? This is likely to be even more confusing - e.g. do I have to set `EMISSION_SERVER_HOME` twice? 

Similarly, the existing instructions already say (on lines 31-32), that you should `cd` into the `<eval_folder>` before copying the conf files over. Why not clarify that instead of saying that you need to copy files over again?

README.md
humbleOldSage: I'll correct it right away.

humbleOldSage:
Rather than repeating the instructions, I have now included a check to ensure that right path was added in the environment variables.

shankari:
@humbleOldSage this looks good, but incomplete. What about highlighting that you have the copy the `conf` directory **after** `cd` into the correct directory?

humbleOldSage:
Added an example of what an '<eval_folder>' could be since this was the part where I felt a bit confused and created my own empty eval folder.

README.md
```diff
@@ -13,19 +13,25 @@ it needs to be included while running them.
...
+    and ensure ```ENV_SERVER_HOME``` is present in the list and has the right path (as mentioned above).
+    
 1. If you haven't setup before, set up the evaluation system
 
     ```
     $ source setup.sh
     ```
-
-1. If you have, activate
+ 1. If you have, activate
```
shankari, line 27: Extraneous whitespace, please fix.

shankari (changes requested):
Almost there! I will squash merge this this experimental change has a kind of messy set of commits. This means that your master will have what appear to be unresolved commits. Please handle this properly (e.g. by deleting master/deleting the fork) and make sure to submit from non-master branches in the future.

humbleOldSage:
I'll be careful with the branches moving forward .

shankari (approved):

shankari:
Congratulations on your first merged PR! 🎉

Issue #35, opened 2023-07-24
Clean up the TRB_label_assist code


shankari:
This will:
- help us modify it and make it better if we want to experiment with different algorithms, including trajectory matching
- allow us to integrate into our production system
- make sure that it is reproducible before the related paper gets published

The code is in: https://github.com/e-mission/e-mission-eval-private-data/tree/master/TRB_label_assist
It works, but it requires us to use a custom fork of the e-mission-server repo to work

We should see the changes between the custom fork and the current e-mission-server master/gis branch
See if the changes are still necessary or whether there are implementations in e-mission-server that have superceded them
If they have not been superceded, we need to incorporate them in e-mission-server

So at the end, this should be reproducible against a core e-mission-server repo

Stretch goal: change this repo to work with docker containers that are built with the base e-mission-server container so that people don't have to install e-mission-server and set the EMISSION_SERVER_PATH and all the funky setup steps.

shankari:
The current implementation of the mode building part of label assist code is at:
https://github.com/e-mission/e-mission-server/tree/master/emission/analysis/modelling/trip_model

There are previous implementations in https://github.com/e-mission/e-mission-server/tree/master/emission/analysis/modelling/tour_model and https://github.com/e-mission/e-mission-server/tree/master/emission/analysis/modelling/tour_model_first_only

- tour model was written by a summer intern and is very bad code
- tour_model_first_only was a last minute hack by me to salvage some of the interns code and get a feature working in production
- trip_model is a rewrite of tour_model_first_only to be more modular

The custom branch uses some code from `tour_model` or `tour_model_first_only` since the TRB_label_assist work was being done in parallel with the trip_model rewrite.

So we should:
- see where that functionality is in trip_model
- if it exists, replace its usage in TRB_label_assist with the implementation from trip_model
- if it doesn't exist, we add it to trip_model and then replace its usage in TRB_label_assist

**Eventually** we should take the new and improved model(s) from TRB label assist and put the into `trip_model`

For a high level summary of what the label assist model does, please see this poster:
https://www.osti.gov/biblio/1922617

humbleOldSage:
Two files, Clustering.py and models.py needs to be updated for us to port the entire code to just our  main repo. 
I had two questions: 

[...]

PR #37, opened 2023-08-11
Update clustering.py

humbleOldSage:
Changes in `clustering.py` file to shift dependency from hlu09's  `tour_model_extended` to main branch's `trip_model`. Still need to change type of data being passed to fit function for this to work. Marked with a TODO.  Explained in detail at https://github.com/e-mission/e-mission-eval-private-data/issues/35#issuecomment-1674290156


TRB_label_assist/clustering.py
```diff
@@ -101,15 +102,17 @@ def add_loc_clusters(
...
+            model_config = {
+                "metric": "od_similarity",
+                "similarity_threshold_meters": r,  # meters,
+                "apply_cutoff": False,
+                "incremental_evaluation": False
+            }    
+
+            sim_model = eamtg.GreedySimilarityBinning(model_config)
+            sim_model.fit(loc_df)   # TODO: Change the type of data being passed to 
+            labels = sim_model.tripLabel
```
shankari, line 115: why isn't this just `labels = sim_model.keys()`

humbleOldSage: do you mean `sim_model.bins.keys()`? 

`sim_model.bins.keys()` would return just the ids of the bins.  For e.g.  0 , 1 ,2  and so on. That too in any order. What we need is a trip and binNo. mapping, For e.g. like this

Trip 1   -->    bin No. (say 2)
Trip 2.  -->    bin No. (say 3) 
Trip 3   -->   bin No.( say 2)
Trip 4   -->   bin No.( say 1)
Trip 5   -->   bin No.( say 1)
.
.
.

shankari: I wrote this before https://github.com/e-mission/e-mission-server/pull/930#discussion_r1291765346

TRB_label_assist/clustering.py
```diff
@@ -28,9 +29,26 @@
     'mean_shift'
 ]
 
+def CleanEntryTypeData(loc_df,loc_entry):
+
+    """
+    Helps weed out entries from the list of entries who's which were removed from the df using
```
shankari, line 35: nit: remove `who's`

TRB_label_assist/clustering.py
```diff
@@ -28,9 +29,26 @@
     'mean_shift'
 ]
 
+def CleanEntryTypeData(loc_df,loc_entry):
```
shankari, line 32: nit: if this is a function, it should start with a lower case - only classes start with upper case in CamelCase

TRB_label_assist/clustering.py
```diff
@@ -53,6 +71,9 @@ def add_loc_clusters(
         Args:
             loc_df (dataframe): must have columns 'start_lat' and 'start_lon' 
                 or 'end_lat' and 'end_lon'
+            loc_entry ( list of Entry/confirmedTrip): list consisting all entries from the
+                time data was loaded. loc_df was obtained from this by converting to df and 
+                then filtering out labeled trips and expanding user_inputs
```
shankari, line 76: nit: this is inconsistent - if it is a list of confirmed trip entries, then you might want to rename it `trip_entry` instead of `loc_entry`. `loc` generally stands for `location`.

TRB_label_assist/clustering.py
```diff
@@ -28,9 +29,26 @@
...
+    Helps weed out entries from the list of entries who's which were removed from the df using
+    esdtq.filter_labeled_trips() and esdtq.expand_userinputs()
+
+    loc_df : dataframe amde from entry type data
+    loc_entry : the entry type equivalent of loc_df ,
+                which was passed alongside the dataframe while loading the data
+
+    """
+
+    ids_in_df=set(loc_df['_id'])
```
shankari, line 44: why do you need to use `set`? The `_id` entries are database entries and are guaranteed to be unique.
(as an aside, even if they are not unique, I believe `entry['_id'] in ids_in_df` will still work - it doesn't require a unique list IIRC)

TRB_label_assist/clustering.py
```diff
@@ -28,9 +29,26 @@
...
+    filtered_loc_entry= [entry for entry in loc_entry if entry['_id'] in ids_in_df ]
```
shankari, line 45: have you run this code? Don't you need an else for the if?
You might want to try `filter` instead

TRB_label_assist/clustering.py
```diff
@@ -98,19 +119,23 @@ def add_loc_clusters(
...
+        cleaned_loc_entry= CleanEntryTypeData(loc_df,loc_entry)
+
         for r in radii:
             # this is using a modified Similarity class that bins start/end
             # points separately before creating trip-level bins
-            sim_model = eamts.Similarity(loc_df,
-                                         radius_start=r,
-                                         radius_end=r,
-                                         shouldFilter=False,
-                                         cutoff=False)
```
shankari, line 108: I don't see these in the new model config. Are the default set to these values?

shankari (changes requested):
Have you run the code?
Please indicate "testing done".
Do you get the same graphs as the paper?
Note that the check is not "do I get a graph", it is "do I get *the same* graph"

TRB_label_assist/clustering.py
humbleOldSage: used set just for faster lookup (at least on average) as compared to list. Even without it, this will work

humbleOldSage:
> Have you run the code? 

Yes 

>Do you get the same graphs as the paper? 

Yes 

>Please indicate "testing done".

Ongoing. I'll match and compare labels generated by both custom branch and master branch. This will verify that master branch and custom branch are functioning similarly.

TRB_label_assist/clustering.py
humbleOldSage: Yes.

humbleOldSage:
Do you get the same graphs as the paper? 

They differ. Let me check why this is happening.

humbleOldSage:
Tested. This is running with no errors. 
Can confirm this generates the same results.

TRB_label_assist/clustering.py
```diff
@@ -130,6 +131,7 @@ def add_loc_clusters(
                 "metric": "od_similarity",
                 "similarity_threshold_meters": r,  # meters,
                 "apply_cutoff": False,
+                "clustering_way": clustering_way,
```
shankari, line 134: Some additional context in the commit message or as a comment would be helpful.
Does the existing model already support `clustering_way`, for example?
What are the valid options?

`commit -m` is useful when you are learning git, not so much for long-term projects

shankari (changes requested):
this particular change seems fine if it works.
- Turned out to be pretty simple after all?!
- I would like to see more information in the PR issue that it works (screenshots, information about the model indicating that it works)
- is this the only notebook that is affected by the change? I know that we have a notebook which generates the performance (accuracy/F-score) of various algorithms. I would expect that it would also need to be changed...

humbleOldSage:
> * is this the only notebook that is affected by the change? I know that we have a notebook which generates the performance (accuracy/F-score) of various algorithms. I would expect that it would also need to be changed...

This PR has changes for Comment#1  in https://github.com/e-mission/e-mission-eval-private-data/issues/35#issuecomment-1670358021. Even though there are 2 more places to changes, this PR covers majority of changes.

TRB_label_assist/clustering.py
humbleOldSage: Yeah sure. I was going to write that after I raised https://github.com/e-mission/e-mission-server/pull/933 which supports this.

humbleOldSage: >Does the existing model already support `clustering_way`, for example? What are the valid options?

No This config parameter is introduced in this PR. 
The valid options are :
`origin`
`destination`
`origin-destination`.

Previously, the entire model was using `origin-destination` as the default and only option for similarity.

TRB_label_assist/clustering.py
humbleOldSage: fixed.

TRB_label_assist/clustering.py
humbleOldSage: fixed.

TRB_label_assist/clustering.py
humbleOldSage: fixed.

humbleOldSage:
> * I would like to see more information in the PR issue that it works (screenshots, information about the model indicating that it works)

Screenshot from the latest run, so no errors.

screencapture-localhost-8888-notebooks-clustering-examples-ipynb-2023-08-20-16_14_14

humbleOldSage:
Left is current result. Right is from research paper.
Suburban 50m. 
<p float="left">





</p>

Suburban 100m

<p float="left">


</p>

Suburban 150m

<p float="left">






</p>

humbleOldSage:
Left is current result. Right is from research paper.
College 50m.

<p float="left">



</p>


College 100m
<p float="left">




</p>



College 150m
<p float="left">



</p>

TRB_label_assist/SVM_decision_boundaries.ipynb
humbleOldSage, line 1: Before (Left) and After (Right) visualizations for SVM_decision_boundaries: 

College Clusters 150m :

<p float="left">


</p>


Suburb clusters 100 m :

<p float="left">


</p>

Suburb Clusters 150m :

<p float="left">


</p>

TRB_label_assist/clustering_examples.ipynb
```diff
@@ -57,7 +57,6 @@
     "uuids = [suburban_uuid, college_campus_uuid]\n",
     "confirmed_trip_df_map = {}\n",
     "labeled_trip_df_map = {}\n",
-    "EntryList=[]\n",
```
humbleOldSage: Just reversed a past unnecessary change here

TRB_label_assist/generate_figs_for_poster.ipynb
humbleOldSage, line 1: Before (left) and after (right) comparisons :
<p float="left">


</p>

<p float="left">


</p>

<p float="left">


</p>

<p float="left">


</p>

<p float="left">


</p>

humbleOldSage: The last 3-4 cells in this notebook won't run. Particularly because `Shankar_df` named variable is not defined.

TRB_label_assist/models.py
```diff
@@ -315,17 +316,23 @@ def fit(self, train_df):
         if len(self.train_df) == 0:
             # i.e. no valid trips after removing all nans
             raise Exception('no valid trips; nothing to fit')
-
+        
+        model_config = {
```
humbleOldSage, line 320: The changes her, from line 319 till 335 are the same changes as clustering.py file. They remove dependency from `eamts` in `tour_model_extended` in custom branch.

TRB_label_assist/models.py
```diff
@@ -880,13 +887,19 @@ def set_params(self, params):
 
         return self
 
-    def fit(self, train_df):
+    def fit(self, train_df,ct_entry):
```
humbleOldSage, line 890: `ct_entry` is the `Entry` type data kept aside when loading data and converting to initial dataframe. 

PS: `train_df` , the other parameter, is derived from  this initial dataframe.

TRB_label_assist/models.py
```diff
@@ -880,13 +887,19 @@ def set_params(self, params):
...
         # fit clustering model
-        self.end_cluster_model.fit(train_df)
+        if self.__class__.__name__ == 'RefactoredNaiveCluster':
```
humbleOldSage, line 892: passing the `Entry' type data to `fit` function of `RefactoredNaiveCluster`. Since there are other models as well ( with their own fit), changing the `fit` without if-else would mean we would have change fit functions definitions of  all involved models and pass `ct_entry` to these even when they don't  need it.

TRB_label_assist/performance_eval.py
```diff
@@ -171,7 +172,10 @@ def cross_val_predict(model,
 
         # train the model
         logging.info("About to fit the model %s" % model)
-        model_.fit(train_trips)
+        if model_.__class__.__name__ == 'ClusterExtrapolationClassifier' :
```
humbleOldSage, line 175: `ClusterExtrapolationClassifier` has dependence on `RefactoredNaiveCluster` ( as mentioned in comments  https://github.com/e-mission/e-mission-eval-private-data/pull/37/commits/e2448c5f584e483e089077062b0c74a2e178bf7e#r1302423250 ). So, the `Entry` type data is passed from `ClusterExtrapolationClassifier` to `RefactoredNaiveCluster` . if-else used because there are other model's`fit` functions that have no requirement of `Entry` data.

humbleOldSage:
4 notebooks, 

`generate_figs_for_poster.ipynb`
`SVM_decision_boundaries.ipynb`
`cluster_performance.ipynb`
`clustering_example.ipynb`

are now working correctly.

TRB_label_assist/models.py
shankari: @humbleOldSage when I said 

> Some additional context in the commit message or as a comment would be helpful.

I meant "source code comments"
https://github.com/e-mission/e-mission-eval-private-data/pull/37#discussion_r1299412247

While this is better than nothing - it requires people to use git blame to find the commit, then find the related PR and scroll down to find these comments. And since we will be resolving them,  they would all be hidden, so they would not be searchable anyway.

TRB_label_assist/models.py
shankari: I see what you say. However, in terms of clean interfaces, I think it would actually be better to have all the `fit` methods take in the same set of parameters, so that we can run them interchangeably and compare the results. 

Think of `sklearn` - does it have separate interfaces for each type of ML model? No.
One of its many great strengths is that it provides a standard interface, so you can write something like 

```
for model in list_of_models:
    model.fit(X, y)
    ...
```

That is what we are going for here. I know that you said that you haven't run `performance_eval.py` yet, but you might want to at least look at what it expects. 

It essentially does something very similar

```
        model_ = model()
        if model_params is not None:
            model_.set_params(model_params)
        train_trips = user_df.iloc[train_idx]
        test_trips = user_df.iloc[test_idx]

        # train the model
        logging.info("About to fit the model %s" % model)
        model_.fit(train_trips)
```

I thought that Hannah did a great job of putting a nice, generic interface on top of all the models that we compare in the paper

TRB_label_assist/performance_eval.py
shankari: Please see my comment above wrt unified interfaces.

TRB_label_assist/clustering.py
shankari: But, as I said, these are guaranteed to be unique. So wrapping them in a set is unnecessary overhead.

TRB_label_assist/clustering.py
shankari: where is `shouldFilter`?

shankari (changes requested):
please see my responses to your previous comments as well
also, please hold off on resolving comments - I will resolve them once I have read your response.

TRB_label_assist/models.py
humbleOldSage: yeah that makes sense.

humbleOldSage:
These are the results I got from running `get_performance_for_poster.ipynb`
Screen Shot 2023-08-30 at 3 36 45 PM

TRB_label_assist/clustering.py
```diff
@@ -16,8 +16,9 @@
 # our imports
 # NOTE: this requires changing the branch of e-mission-server to
 # eval-private-data-compatibility
-import emission.analysis.modelling.tour_model_extended.similarity as eamts
+# import emission.analysis.modelling.tour_model_extended.similarity as eamts
```
shankari, line 19: please remove instead of commenting

TRB_label_assist/models.py
```diff
@@ -23,7 +23,14 @@
 import emission.analysis.modelling.tour_model_first_only.evaluation_pipeline as ep
```
shankari, line 23: you should remove these imports as well. There is really no reason for us to keep using `tour_model_first_only`. Is is not used in the production server code; it has been replaced by `trip_model` there.

TRB_label_assist/regenerate_classification_performance_results.py
```diff
@@ -15,10 +16,11 @@
 labeled_trip_df_map = {}
 expanded_labeled_trip_df_map = {}
 expanded_all_trip_df_map = {}
+ct_entry={}
 for u in all_users:
     ts = esta.TimeSeries.get_time_series(u)
     ct_df = ts.get_data_df("analysis/confirmed_trip")
-
+    ct_entry[u]=eamtr._get_training_data(u,None)
```
shankari, line 23: this will read twice from the database. It seems like you should be able to get_training_data and then use `_to_data_df` to convert the entries into a dataframe.

shankari (changes requested):

TRB_label_assist/models.py
```diff
@@ -367,7 +382,7 @@ def _match(self, trip, bin, loc_type):
             copied from the Similarity class on the e-mission-server. 
         """
         for t_idx in bin:
-            trip_in_bin = self.train_df.iloc[t_idx]
+            trip_in_bin = self.train_df.iloc[int(t_idx)]
```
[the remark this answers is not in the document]
humbleOldSage: the bins in `bin` are stored as list of strings, for e.g. ['1','2','3'....]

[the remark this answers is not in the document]
shankari: so how did it work before?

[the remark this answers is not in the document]
humbleOldSage: the `tour_model` generated bins in the form [1,2,3 ...] . 
 previously , this is how we got bins from the model:
`labels = sim_model.data_df[loc_type + '_bin'].to_list()`.    ( Before changes Line 112 in [clustering.py ](https://github.com/e-mission/e-mission-eval-private-data/pull/37/files) )


The 'trip_model' saves it as mentioned in my previous comment above. 
this is how we got bins from the model now :
 'labels = [int(l) for l in sim_model.tripLabels]'.  ( After changes Line 140 in [clustering.py ](https://github.com/e-mission/e-mission-eval-private-data/pull/37/files) )

TRB_label_assist/clustering.py
humbleOldSage: did.

TRB_label_assist/models.py
humbleOldSage: done.

TRB_label_assist/regenerate_classification_performance_results.py
humbleOldSage: fixed.

humbleOldSage:
PARTIALLY TESTED -- I have tested all the notebook except the one that'll take 2 days to run. It's been running for a day now .As soon as that's done. I'll update here.

TRB_label_assist/SVM_decision_boundaries.ipynb
shankari, line 1: It is impossible to review this file because of the formatting changes.
Almost every single line shows a diff.
Please ensure that only the lines that have changed show up in the diff to avoid polluting the commit history..

TRB_label_assist/classification_performance.ipynb
shankari, line 1: Ditto

TRB_label_assist/cluster_performance.ipynb
shankari, line 1: Ditto

TRB_label_assist/clustering.py
shankari: This has not yet been resolved

TRB_label_assist/clustering_examples.ipynb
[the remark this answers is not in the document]
shankari: Similar to previous extraneous whitespace changes, please ensure that only your changes are reflected.

TRB_label_assist/first_trial_results/cv results DBSCAN+SVM (destination).csv
shankari, line 1: Ditto

TRB_label_assist/regenerate_classification_performance_results.py
```diff
@@ -15,10 +16,11 @@
 labeled_trip_df_map = {}
 expanded_labeled_trip_df_map = {}
 expanded_all_trip_df_map = {}
+ct_entry={}
 for u in all_users:
     ts = esta.TimeSeries.get_time_series(u)
-    ct_df = ts.get_data_df("analysis/confirmed_trip")
-
+    ct_entry[u]=eamtr._get_training_data(u,None)    
+    ct_df = ts.to_data_df("analysis/confirmed_trip",itertools.chain(ct_entry[u]))
```
shankari, line 23: why do you need this change?

TRB_label_assist/models.py
```diff
@@ -648,29 +662,37 @@ def set_params(self, params):
 
         return self
 
-    def fit(self, train_df):
+    def fit(self, train_df,ct_entry=None):
```
shankari, line 665: in general, if your variable is a list, it is good to reflect that in the name to make it easier to  read.
This sounds like you are passing in a single entry.

TRB_label_assist/models.py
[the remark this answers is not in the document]
shankari: so the difference here is that we are using strings instead of ints?

shankari (changes requested):
Several of my previous comments were not addressed.
In particular, how are you running the notebooks? They are unreviewable right now due to the extraneous whitespace changes.

TRB_label_assist/regenerate_classification_performance_results.py
humbleOldSage: This'll prevent two reads from the database. This is done in response to  https://github.com/e-mission/e-mission-eval-private-data/pull/37#discussion_r1312331029 . We need the data in df as well as in entry format.

TRB_label_assist/first_trial_results/cv results DBSCAN+SVM (destination).csv
humbleOldSage: handled.

TRB_label_assist/SVM_decision_boundaries.ipynb
humbleOldSage: Since I was running the .ipynbs in browser, it changed cells indexed every time a cell is run, this causes .json to be changed. I switched to VS code, which kinda fixed this issues, except that it was adding  trailing or leading extra spaces in certain cells and removing trailing spaces from others. I fixed this manual by opening .ipynb in text editor.

Now, hopefully this is fixed.

TRB_label_assist/models.py
humbleOldSage: addressed. will follow in next commit.

TRB_label_assist/models.py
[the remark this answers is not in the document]
humbleOldSage: the way GreedySimilarityBinning was implemented, these are stored as strings. But converted to ints when required.

TRB_label_assist/SVM_decision_boundaries.ipynb
```diff
@@ -60,10 +62,12 @@
                 "uuids = [suburban_uuid, college_campus_uuid]\n",
                 "confirmed_trip_df_map = {}\n",
                 "labeled_trip_df_map = {}\n",
+                "ct_entry={}\n",
                 "expanded_trip_df_map = {}\n",
                 "for u in uuids:\n",
                 "    ts = esta.TimeSeries.get_time_series(u)\n",
-                "    ct_df = ts.get_data_df(\"analysis/confirmed_trip\")\n",
+                "    ct_entry[u]=eamtr._get_training_data(u,None)\n",
+                "    ct_df = ts.to_data_df(\"analysis/confirmed_trip\",itertools.chain(ct_entry[u]))\n",
```
shankari, line 70: why do we need `itertools.chain` here? `find_entries` already returns a list of entries, IIRC
Please add code comment since the related blame does not have any explanation

TRB_label_assist/SVM_decision_boundaries.ipynb
```diff
@@ -264,9 +270,9 @@
     ],
     "metadata": {
         "kernelspec": {
-            "display_name": "Python 3",
+            "display_name": "emission-private-eval",
```
shankari, line 273:
```suggestion
            "display_name": "Python 3",
```

TRB_label_assist/SVM_decision_boundaries.ipynb
```diff
@@ -264,9 +270,9 @@
...
             "language": "python",
-            "name": "python3"
+            "name": "emission-private-eval"
```
shankari, line 275:
```suggestion
            "name": "python3"
```

TRB_label_assist/SVM_decision_boundaries.ipynb
```diff
@@ -278,7 +284,7 @@
             "name": "python",
             "nbconvert_exporter": "python",
             "pygments_lexer": "ipython3",
-            "version": "3.9.6"
+            "version": "3.9.16"
```
shankari, line 287:
```suggestion
            "version": "3.9.6"
```

TRB_label_assist/classification_performance.ipynb
```diff
@@ -49,10 +49,11 @@
                 "labeled_trip_df_map = {}\n",
                 "expanded_labeled_trip_df_map = {}\n",
                 "expanded_all_trip_df_map = {}\n",
+                "ct_entry={}\n",
                 "for u in all_users:\n",
                 "    ts = esta.TimeSeries.get_time_series(u)\n",
-                "    ct_df = ts.get_data_df(\"analysis/confirmed_trip\")\n",
-                "\n",
+                "    ct_entry[u]=eamtr._get_training_data(u,None)\n",
+                "    ct_df = ts.to_data_df(\"analysis/confirmed_trip\",itertools.chain(ct_entry[u]))\n",
```
shankari, line 56: Ditto

TRB_label_assist/clustering.py
shankari: This has not yet been answered

TRB_label_assist/models.py
shankari: This has not yet been addressed.

TRB_label_assist/performance_eval.py
```diff
@@ -274,6 +278,7 @@ def cv_for_all_algs(uuid_list,
                     min_samples=False,
                     raise_errors=False):
     cv_results = {}
+    pathlib.Path('first_trial_results').mkdir(parents=True,exist_ok=True) #needed first time
```
shankari, line 281: would have been good to have included this change in the description of the related commit

https://github.com/humbleOldSage/e-mission-eval-private-data/commit/a34836faf364f9548d3ab5373c8efd0d1e729a53

TRB_label_assist/models.py
```diff
@@ -1049,7 +1071,7 @@ class EnsembleClassifier(TripClassifier, metaclass=ABCMeta):
     replaced_predictor = NotImplemented
 
     # required methods
-    def fit(self, train_df):
+    def fit(self, train_df,ct_entry=None):
```
shankari, line 1074: I would rename to `unused` to clarify that we don't use `ct_entry` here, or add a code comment if you want to use a named parameter instead of a positional parameter.

```suggestion
    def fit(self, train_df,unused=None):
```

TRB_label_assist/models.py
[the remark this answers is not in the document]
shankari: I am not going to hold up the merge for this, but I don't understand why we can't just convert from strings to ints one time right after return instead of converting for each usage.

TRB_label_assist/models.py
```diff
@@ -159,11 +165,12 @@ def fit_predict(self, train_df):
 class TripClassifier(SetupMixin, metaclass=ABCMeta):
 
     @abstractmethod
-    def fit(self, train_df):
+    def fit(self, train_df,ct_entry=None):
```
shankari, line 168:
```suggestion
    def fit(self, train_df,unused=None):
```

TRB_label_assist/models.py
```diff
@@ -116,11 +121,12 @@ class Cluster(SetupMixin, metaclass=ABCMeta):
     """ blueprint for clustering models. """
 
     @abstractmethod
-    def fit(self, train_df):
+    def fit(self, train_df,ct_entry=None):
```
shankari, line 124:
```suggestion
    def fit(self, train_df,unused=None):
```

shankari (changes requested):
Almost there. Only a few small cleanups and clarifications...

shankari:
The comments here have not been addressed, this is still a draft.
https://github.com/e-mission/e-mission-eval-private-data/issues/35 is not complete

TRB_label_assist/models.py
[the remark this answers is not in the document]
humbleOldSage: Sure, we can. These bins are stored in the `self.bins` variable . We access these bins in the predict function. 

So, we can change the keys from string to int right before we begin our prediction.

TRB_label_assist/SVM_decision_boundaries.ipynb
humbleOldSage: we do not. Following the code trail, I felt it would be needed, but the code runs fine without. REMOVED.

TRB_label_assist/SVM_decision_boundaries.ipynb
humbleOldSage: done.

TRB_label_assist/SVM_decision_boundaries.ipynb
humbleOldSage: done.

TRB_label_assist/classification_performance.ipynb
humbleOldSage: we do not need this. Following the code trail, I felt it would be needed, but the code runs fine without. REMOVED

TRB_label_assist/clustering.py
humbleOldSage: Added back. Was removed by mistake.

TRB_label_assist/models.py
humbleOldSage: Commited.

TRB_label_assist/regenerate_classification_performance_results.py
humbleOldSage: Removed the itertools part.

humbleOldSage:
Since this is partially tested, I'll keep the PR as draft, as soon as I have completed the final testing, I'll mark it as ready to merge.

TRB_label_assist/SVM_decision_boundaries.ipynb
```diff
@@ -270,9 +269,9 @@
     ],
     "metadata": {
         "kernelspec": {
-            "display_name": "emission-private-eval",
```
shankari: why did these names change? Are you not running this notebook in the correct environment?

TRB_label_assist/SVM_decision_boundaries.ipynb
```diff
@@ -284,7 +283,7 @@
             "name": "python",
             "nbconvert_exporter": "python",
             "pygments_lexer": "ipython3",
-            "version": "3.9.16"
```
shankari: Ditto

TRB_label_assist/models.py
```diff
@@ -121,13 +121,13 @@ class Cluster(SetupMixin, metaclass=ABCMeta):
...
     @abstractmethod
-    def fit(self, train_df,ct_entry=None):
+    def fit(self, train_df,unused=None):
         """ Fit the clustering algorithm.  
         
             Args: 
                 train_df (DataFrame): dataframe of labeled trips
-                ct_entry (List) : A list of Entry type of labeled and unlabeled trips 
-            
+                unused (List) : A list of Entry type of labeled and unlabeled trips which is not used in current function.
```
shankari, line 129: why is this `unused` in the abstract class? If it is never used, in any subclass, it can just be removed completely. But it is used, so it should stay and the comments should be updated. Same message throughout the changes in this file.

TRB_label_assist/models.py
```diff
@@ -165,13 +165,13 @@ def fit_predict(self, train_df):
...
     @abstractmethod
-    def fit(self, train_df,ct_entry=None):
+    def fit(self, train_df,unused=None):
         """ Fit a classification model.  
         
             Args: 
                 train_df (DataFrame): dataframe of labeled trips
-                ct_entry (List) : A list of Entry type of labeled and unlabeled trips 
-            
+                unused (List) : A list of Entry type of labeled and unlabeled trips which is not used in current function.
```
shankari, line 173: it makes sense to have it `unused` here

TRB_label_assist/models.py
```diff
@@ -300,7 +300,7 @@ def set_params(self, params):
 
         return self
 
-    def fit(self, train_df,ct_entry=None):
```
shankari: and here `train_df` should be unused

TRB_label_assist/models.py
```diff
@@ -351,6 +351,7 @@ def predict(self, test_df):
         elif self.loc_type == 'end':
             bins = self.sim_model.bins
 
+        bins = {int(key):value for key,value in bins.items()}
```
shankari, line 354:
```suggestion
        # This looks weird but works
        # >>> x = [(1, 'a'), (2, 'b'), (3, 'c')]
        # >>> {int(key):value for key,value in x}
        # {1: 'a', 2: 'b', 3: 'c'}
        bins = {int(key):value for key,value in bins.items()}
```

shankari (changes requested):
Almost done, just a few minor changes

TRB_label_assist/models.py
humbleOldSage:
```
        # bins = { '1': [ .. ....], 
        #          '2': [......], 
        #          '3': [.......] ....}
        # the  code converts this to 
        # bins = { 1: [ .. ....], 
        #          2: [......], 
        #          3: [.......] ....}
        ```

File: TRB_label_assist/SVM_decision_boundaries.ipynb (hunk 29 above)
Comment (humbleOldSage) [reply TRB_label_assist/SVM_decision_boundaries.ipynb]: the environments is correct.However, it shows up as `emission-private-eval` ( and not Python3) on the json of the notebook as soon as I run even 1 cell. This might have slipped passed me in one of the previous commits as I usually reset this before committing.

File: TRB_label_assist/SVM_decision_boundaries.ipynb (hunk 30 above)
Comment (humbleOldSage) [reply TRB_label_assist/SVM_decision_boundaries.ipynb]: different python version than the one used by person who committed this previously. This shows up as 3.9.6 on the json of the notebook as soon as I run even 1 cell. I usually reset this before commits to not create unnecessary history. This might have slipped passed me in one of the previous commits.

**Comment by humbleOldSage** [comment]
Not tested. Needs Testing.

File: TRB_label_assist/models.py (hunk 34 above)
Comment (shankari) [reply TRB_label_assist/models.py:362]: right. I'm just suggesting that you add a comment there indicating that it works since it looks weird to create a dict without a list (I understand that it works because of the iterable, but let's make life easier for others in the future).

File: TRB_label_assist/SVM_decision_boundaries.ipynb (hunk 29 above)
Comment (shankari) [reply TRB_label_assist/SVM_decision_boundaries.ipynb]: And now it is `Python3` instead of `Python 3`. Argggh!!

@humbleOldSage **Can you please revert these as well?**

We should then do a separate one-time change to make the metadata current, with **no other changes**, similar to 
https://github.com/e-mission/e-mission-phone/commit/bb73676c7245b68b9a8784204b68d71ba0af5267

File: TRB_label_assist/clustering_examples.ipynb [hunk 35]
```diff
@@ -26,12 +26,13 @@
                 "%autoreload 2\n",
                 "\n",
                 "from uuid import UUID\n",
-                "\n",
+                "import itertools\n",
```
shankari, line 29: extraneous import

TRB_label_assist/clustering_examples.ipynb
```diff
@@ -26,12 +26,13 @@
...
                 "import emission.storage.timeseries.abstract_timeseries as esta\n",
                 "import emission.storage.decorations.trip_queries as esdtq\n",
                 "import emission.core.get_database as edb\n",
-                "\n",
-                "import mapping"
+                "import emission.analysis.modelling.trip_model.run_model as eamtr\n",
+                "import mapping\n",
+                "import emission.analysis.modelling as eam"
```
shankari, line 35: what do we use this for? we typically import the leaf modules and not a package

TRB_label_assist/generate_figs_for_poster.ipynb
```diff
@@ -29,15 +29,15 @@
                 "import numpy as np\n",
                 "import matplotlib.pyplot as plt\n",
                 "import matplotlib\n",
-                "\n",
+                "import itertools\n",
```
shankari, line 32: Ditto

TRB_label_assist/get_performance_for_poster.ipynb
```diff
@@ -22,9 +22,10 @@
                 "from uuid import UUID\n",
                 "\n",
                 "import matplotlib.pyplot as plt\n",
-                "\n",
+                "import itertools\n",
```
shankari, line 25: Ditto

shankari (changes requested):
Even smaller cleanups

TRB_label_assist/SVM_decision_boundaries.ipynb
shankari: In general, when you respond like this, please clarify what (if anything) you have done to address the comment.
I can then verify it and resolve.

humbleOldSage:
`generate_figs_from_poster.ipynb` :

plot after latest testing

Screenshot 2023-11-16 at 12 01 08 PM

 snap from the research paper :
Screenshot 2023-11-16 at 3 40 13 PM

plot after latest testing
Screenshot 2023-11-16 at 12 01 40 PM

 snaps from the research paper :
Screenshot 2023-11-16 at 3 43 06 PM

humbleOldSage:
`generate_figs_for_poster.ipynb` :

On the left are Plots after current testing, on the right are images from runs of notebook with @hlu109  custom branch :

### naive fixed-width clustering from the first user's data

150m 




</p>

50m

<p float="left">




</p>

100m
<p float="left">




</p>

### DBSCAN without SVM: home cluster with a blue cluster to the south that was merged in

<p float="left">





</p>


### DBSCAN + SVM: home cluster and blue cluster to the south have been separated

<p float="left">





</p>

humbleOldSage:
`Clustering_example.ipynb`

Left is current test result. Right is from research paper.
Suburban 50m. 
<p float="left">





</p>

Suburban 100m

<p float="left">



</p>

Suburban 150m

<p float="left">






</p>





Left is current result. Right is from research paper.
College 50m.

<p float="left">



</p>


College 100m
<p float="left">




</p>



College 150m
<p float="left">



</p>

humbleOldSage:
`SVM_decision_boundary.ipynb` :

On the left are plots from current test, on the right are plots from old runs :

<p float="left">





</p>
<p float="left">




</p>

<p float="left">







</p>

humbleOldSage:
`get_cluster_performance.ipynb` :

For each pair, top one is the result of current test, bottom one is result from older runs :

output2

Screen Shot 2023-08-22 at 1 05 03 AM

output3

Screen Shot 2023-08-22 at 1 05 34 AM



output4


Screen Shot 2023-08-22 at 1 05 44 AM


output5


Screen Shot 2023-08-22 at 1 05 52 AM

TRB_label_assist/clustering_examples.ipynb
humbleOldSage: we use `find_plot_clusters` from this file. That's our custom .py file. 

We have used mapping and clustering in some other files as well:

```
import mapping
import clustering
```

TRB_label_assist/generate_figs_for_poster.ipynb
humbleOldSage: done. This was not tested till that point.

TRB_label_assist/clustering_examples.ipynb
humbleOldSage: done. This was not tested till that point.

humbleOldSage:
All model results :

Screenshot 2023-11-16 at 5 53 00 PM

TRB_label_assist/clustering_examples.ipynb
shankari: 1. which other files?
2. My comment was primarily around `import emission.analysis.modelling as eam` (where `emission.analysis.modelling` is the package) elsewhere. We typically import individual modules.

TRB_label_assist/SVM_decision_boundaries.ipynb
shankari: This still shows as a change

TRB_label_assist/SVM_decision_boundaries.ipynb
shankari: and so does this

TRB_label_assist/SVM_decision_boundaries.ipynb
shankari: This has still not been resolved

TRB_label_assist/clustering_examples.ipynb
```diff
@@ -60,9 +60,11 @@
                 "confirmed_trip_df_map = {}\n",
                 "labeled_trip_df_map = {}\n",
                 "expanded_trip_df_map = {}\n",
+                "ct_entry={}\n",
                 "for u in uuids:\n",
                 "    ts = esta.TimeSeries.get_time_series(u)\n",
-                "    ct_df = ts.get_data_df(\"analysis/confirmed_trip\")\n",
+                "    ct_entry[u]=eamtr._get_training_data(u,None)    \n",
+                "    ct_df = ts.to_data_df(\"analysis/confirmed_trip\",itertools.chain(ct_entry[u]))    \n",
```
shankari, line 67: why do you still use `itertools` here?

TRB_label_assist/models.py
```diff
@@ -880,13 +922,13 @@ def set_params(self, params):
 
         return self
 
-    def fit(self, train_df):
+    def fit(self, train_df,train_entry_list=None):
```
shankari, line 925: shouldn't this be `unused` since we were not passing in the entries earlier?

shankari (commented):
@humbleOldSage I don't think you have addressed several of the prior review comments. They are fairly simple, so you might have just missed them - please check the review history carefully.

Please make sure that all comments are addressed before marking as ready for review.

TRB_label_assist/models.py
```diff
@@ -378,13 +405,19 @@ def _distance_helper(self, tripa, tripb, loc_type):
         
             copied from the Similarity class on the e-mission-server.
```
shankari, line 406: Now that we are no longer on a custom branch, can't the distance calculation in `e-mission-server` be re-used here? Why do we need copy-pasted code? I am generally again copy-pasting to support DRY. Since this change has already had multiple revisions, I am OK with deferring this to the next PR, but I want to make sure that it is not forgotten.

TRB_label_assist/models.py
humbleOldSage: In case when `start_cluster_model` is `RefractoredNaiveCluster` ( which is based on greedy similarity binning), we use the list of entry type data while fitting. However if `start_cluster_model` is DBSCAN, train_df is used. This is based on the value of `self.alg` parameter passed.

TRB_label_assist/models.py
humbleOldSage: We are using the `e-mission-server` :

```
  dist= ecc.calDistance([pta_lon,pta_lat],[ptb_lon,ptb_lat])     
```                              

ecc here is on `e-mission-server` :
```
  import emission.core.common as ecc
```

I'll remove that old comment from Hannah.

TRB_label_assist/clustering_examples.ipynb
humbleOldSage: removed.

TRB_label_assist/clustering_examples.ipynb
humbleOldSage: 1. My understanding was around `mapping.py` and `clustering.py` which are leaf modules.
 
2.`eam` is unnecessary here. We are importing anyways importing `eamtr` in case when we need to run the model. I'll remove it.

TRB_label_assist/models.py
humbleOldSage: I am unable to get what's expected here. I wrote an example there explaining what it does. I am now adding why it works there. Let me know if I missed something.

TRB_label_assist/SVM_decision_boundaries.ipynb
humbleOldSage: Reverted this back to Python 3 using git file history.

TRB_label_assist/SVM_decision_boundaries.ipynb
humbleOldSage: Reverted this back to Python 3 using git file history.

TRB_label_assist/SVM_decision_boundaries.ipynb
humbleOldSage: Reverted this back to Python 3 using git file history.

TRB_label_assist/models.py
humbleOldSage: While predicting in the `greedy_similarity_binning.py` on `e-mission-server`,  the flow goes like :

 `predict` -> `_nearest_bin` ->`similar` ( in `e-mission-server/emission/analysis/modelling/similarity/similarity_metric.py`) ->`similarity` ( in `e-mission-server/emission/analysis/modelling/similarity/od_similarity.py`) -> `ecc.calDistance`

TRB_label_assist/models.py
shankari: @humbleOldSage I have a suggested change right here.
https://github.com/e-mission/e-mission-eval-private-data/pull/37#discussion_r1388696448
suggested 2 weeks ago

All you have to do is to incorporate the suggested change. I am not sure what is not clear here.

TRB_label_assist/models.py
shankari: I think that you have misunderstood the comment. The comment is not about `ecc.calDistance` - it is clear that `calDistance` is from 
This says that `_distanceHelper` is copied from e-mission-server, which is it

```
$ grep -r distance_helper emission/
emission//analysis/modelling/tour_model_first_only_orig/similarity.py:            if not self.distance_helper(a, b):
emission//analysis/modelling/tour_model_first_only_orig/similarity.py:    def distance_helper(self, a, b):
emission//analysis/modelling/tour_model/similarity.py:            if not self.distance_helper(a, b):
emission//analysis/modelling/tour_model/similarity.py:    def distance_helper(self, a, b):
emission//incomplete_tests/TestSimilarity.py:                    self.assertTrue(sim.distance_helper(b,c))
```

However, the implementation does seem to be a bit different, and I don't see a function with this name in trip_model. There must be an equivalent in trip_model.

shankari:
@humbleOldSage two more comments.

TRB_label_assist/models.py
humbleOldSage: Other than `ecc.calDistance`, currently  pre-processing is done  in the form of coordinate extraction in `_distanceHelper`:

```
        #tripa is taken from the test datframe. 
        #tripb is taken from the stored bin list.
        
        pta_lat = tripa[[loc_type + '_lat']]
        pta_lon = tripa[[loc_type + '_lon']]
        if loc_type == 'start':
            ptb_lat = tripb[1]
            ptb_lon = tripb[0]
        elif loc_type == 'end':
            ptb_lat = tripb[3]
            ptb_lon = tripb[2]
```


We do have `extract_features` function ( at `e-mission-server/emission/analysis/modelling/similarity/od_similarity.py`) that extracts latitude and longitude of trips, but it works just for `Entry` type data ( since data frames are not used in `e-mission-server`). 
Here, in this repo, we additionally have `tripa` ( the test trip),which is row of data frame received from the `predict` function. 


>There _must_ be an equivalent in trip_model to calculate distances between trips, which we should reuse here.

For the reason above, there isn't. However, this is  what we can do  to use the `_nearest_bin` function ( `e-mission-server/emission/analysis/modelling/trip_model/greedy_similarity_binning.py`)  which is closest to `_distance_helper` :

1. convert `tripa` ( the test trip) to entry type data  using  `df_row_to_entry` ( in `e-mission-server/emission/storage/timeseries/builtin_timeseries.py`). 
2. Pass this entry to the `_nearest_bin` function.

Let me know if this works, and I'll test this.

shankari: From https://github.com/e-mission/e-mission-eval-private-data/pull/37#discussion_r1398304851

> Since this change has already had multiple revisions, I am OK with deferring this to the next PR, but I want to make sure that it is not forgotten.

I am fine with returning to this later. However, eventually, we should simplify the codebase to either use only dataframes, or use only entries, or, if we are going to support some level of mix and match, have the utility functions support both combinations.

shankari (approved):

shankari:
Squash-merging since this is 21 commits for some fairly simple changes.
@humbleOldSage please account for this while making any future changes.

PR #44, opened 2025-01-10
Notebooks for CanBikeCO Paper Visualizations

Abby-Wheelis:
Ported over the changes from [dashboard #102](https://github.com/e-mission/em-public-dashboard/pull/102) loosely followed [this process](https://stackoverflow.com/questions/41811986/git-move-directory-to-another-repository-while-keeping-the-history) to preserve commit history


Abby-Wheelis:
There was one place where the mapping dictionaries were needed, so pulled those across

shankari:
Got this merged before the paper is published!

